PR #1, opened 2016-08-17
Advancing

fserra:
After @mlubin comment (https://github.com/leethargo/CSIP/pull/17) I realized that I misinterpreted the meaning of `getobjbound`, so the implementation is flawed. For a correct implementation we need to solve Miles' pull request


leethargo:
`CSIP` is now updated. Feel free to fix `getobjbound`.

Should be merge the `advancing` branch then? Or do you want to implement dual LP stuff properly first?

mlubin:
What do you mean by dual LP stuff?

leethargo:
When you use SCIP to solve an LP problem, you don't (easily) get access to data related to the LP solution (dual solution, farkas ray etc.). I don't remember exactly, but there seems to be a workaround where presolve has to be disabled.

So, instead of figuring this out, we added dummy implementations of `getconstrduals` etc. returning `nothing`. This makes it possible to run `linprog` with SCIP.jl, but is rather misleading. I'm fine with not supporting `linprog`, but @fserra suggested we make it work instead.

mlubin:
I'm also fine with not supporting ``linprog`` (Cbc doesn't). That said it wouldn't hurt.

fserra:
yeah I think we can't actually. It seems to be possible if one turns off
presolve... but yeah, for the time being is just not supported

mlubin:
Just leave the methods not implemented then, or have them return something like``error("Duals not available through SCIP")``

PR #9, opened 2016-09-03
enabling qcqpmodel tests doesn't produce errors

fserra:
this needs https://github.com/leethargo/CSIP/pull/21

### DIFF
--- a/src/mpb_interface.jl
+++ b/src/mpb_interface.jl
@@ -219,7 +219,9 @@ setquadobj!(m::SCIPMathProgModel, rowidx, colidx, quadval) =
     error("Not implemented for SCIP.jl")
 
 setquadobjterms!(m::SCIPMathProgModel, rowidx, colidx, quadval) =
-    error("Not implemented for SCIP.jl")
+    _setQuadObj(m, Cint(0), Array{Cint}(0), Array{Cdouble}(0),
+                 Cint(length(rowidx)), convert(Vector{Cint}, rowidx - 1),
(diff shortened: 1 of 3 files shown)


leethargo:
That was easy :smile:

fserra:
why did the test pass? it is using the CSIP version without the `setQuadObj` function...

leethargo:
Maybe there was some old `.so` file being used?

When I try to run the tests with CSIP from `master` branch, there's plenty of errors of this kind:

```
ccall: could not find function CSIPsetQuadObj in library /home/bzfschwa/src/CSIP/lib/libcsip.so
in setquadobjterms! at /home/bzfschwa/.julia/v0.4/SCIP/src/mpb_interface.jl:221
in addQuadratics at /home/bzfschwa/.julia/v0.4/JuMP/src/solvers.jl:444
```

leethargo:
The same kind of errors also occur in Travis, but `FactCheck` seems to eat them all and in the end prints `INFO: SCIP tests passed`, which is why we get the green checkmark.

@mlubin: do you know about this issue with FactCheck and can deal with it?

leethargo:
Instead of using leethargo/CSIP#21, we could also use leethargo/CSIP#22 (after modification).

mlubin:
Looks like the tests are missing a statement like [this](https://github.com/JuliaOpt/JuMP.jl/blob/6fba39ae58963a2dcb20b29001f83d01c9e575ad/test/runtests.jl#L41)

PR #13, opened 2016-09-12
Use BinDeps to build CSIP from source

leethargo:
but require `SCIPOPTDIR` to be set in environment.


leethargo:
I'll try to update Travis, too.

@mlubin: Could you check the `build.jl`? I have no idea about possible version checks or `:unix` annotations that might be missing.

mlubin:
Definitely need to validate the CSIP version because bindeps is useless for updating source depen dependencies.

leethargo:
So we should add a function to CSIP that returns the version?

mlubin:
E.g., https://github.com/JuliaOpt/GLPK.jl/blob/1674755b4f54c9aab64a80f07b2ef5246581e011/deps/build.jl#L21

leethargo:
But this example checks the `VERSION` of julia herself?

mlubin:
That's unrelated, https://github.com/JuliaOpt/GLPK.jl/blob/1674755b4f54c9aab64a80f07b2ef5246581e011/deps/build.jl#L28

leethargo:
I see, thanks.

leethargo:
So, I already specify an exact version of CSIP via the GitHub release URL. The version check will only provide additional consistency, right?

mlubin:
The big issue is what happens when you update the CSIP version. There should be a runtime check in ``__init__`` like [here](https://github.com/JuliaOpt/SCS.jl/blob/94ee319f0f04e1a8b7e334a4d400a6146b235599/src/SCS.jl#L15) otherwise users will get segfaults.

leethargo:
OK, done.

README.md
```diff
@@ -17,3 +17,20 @@ Julia interface to [SCIP](http://scip.zib.de) solver.
...
+
+The SCIP.jl package requires SCIP to be installed.
+
+[Download](http://scip.zib.de/download.php?fname=scipoptsuite-3.2.1.tgz) the
+SCIP Optimization Suite, copy the
+[patched file](http://scip.zib.de/download/bugfixes/scip-3.2.1/Makefile.doit)
+and build the shared library with
+```
+make SHARED=true GMP=false READLINE=false ZLIB=false OPT=opt scipoptlib
+```
```
mlubin: Also need to set ``SCIPOPTDIR``?

leethargo: Of course, I thought I had that already, but it was in the CSIP/Readme.

mlubin:
:+1:

mlubin:
@leethargo, should I register now?

leethargo:
Yes, please go ahead (the tests are all passing).

Issue #14, opened 2016-09-13
document how to set parameters


mlubin:
And how they match up with the SCIP documentation

leethargo:
The parameters match up exactly with the names given in the SCIP documentation (using the `/` separated path syntax). Do you feel this needs furher explanation?

mlubin:
Nope

PR #16, opened 2016-09-13
set parameters in a better way

### LINKED ISSUE #15: float64 parameters
Using `SCIPSolver("limits/gap",1e-5)` leads to

```
ERROR: LoadError: MethodError: `convert` has no method matching convert(::Type{Ptr{Void}}, ::Float64)
This may have arisen from a call to the constructor Ptr{Void}(...),
since type constructors fall back to convert methods.
Closest candidates are:
  call{T}(::Type{T}, ::Any)
  convert{T}(::Type{Ptr{T}}, !Matched::UInt64)
  convert{T}(::Type{Ptr{T}}, !Matched::Int64)
  ...
 in setparams! at /home/mlubin/.julia/v0.4/SCIP/src/params.jl:13
```

Is this the correct usage? Why do you need to convert a `Float64` to a `void*`?

fserra:
it should fix #15

### DIFF
--- a/src/params.jl
+++ b/src/params.jl
@@ -7,10 +7,18 @@
 function setparams!(m::SCIPMathProgModel)
     a = m.options
     for (name, value) in zip(a[1][1:2:end],a[1][2:2:end])
-        if typeof(value) == :ASCIIString
-            value = pointer(value)
-        end
-        _setParameterGeneric(m, pointer(name), convert(Ptr{Void},value))
+        setparameter!(m, name, value)
     end
 end
 
+setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Bool) = _setBoolParam(m, pointer(name), Cint(value))
+function setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Int)
+    if _getParamType(m, pointer(name)) == 1 # integer
+        return _setIntParam(m, pointer(name), Cint(value))
+    else
+        return _setLongintParam(m, pointer(name), Clonglong(value))
+    end
+end
+setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Float64) = _setRealParam(m, pointer(name), Cdouble(value))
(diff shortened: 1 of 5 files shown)


src/params.jl
```diff
@@ -7,10 +7,18 @@
...
+function setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Int)
+    if _getParamType(m, pointer(name)) == 1 # integer
+        return _setIntParam(m, pointer(name), Cint(value))
+    else
+        return _setLongintParam(m, pointer(name), Clonglong(value))
+    end
+end
+setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Float64) = _setRealParam(m, pointer(name), Cdouble(value))
+setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::Char) = _setCharParam(m, pointer(name), Cchar(value))
+setparameter!(m::SCIPMathProgModel, name::ASCIIString, value::ASCIIString) = _setStringParam(m, pointer(name), pointer(value))
```
leethargo: nice, multiple dispatch :+1:

mlubin:
CSIP 0.3.3 thinks that it's 0.3.2: https://github.com/SCIP-Interfaces/CSIP/blob/3e4a43e0a2f0d9c080ce94683853be628ca802b2/src/csip.c#L10

leethargo:
I know, I just saw it. Trying to move the tag now.

leethargo:
I put a stub related to #14, mostly to retrigger the Travis build.

mlubin:
I'm confused how this is passing since CSIP is still at 0.3.2.

leethargo:
What do you mean with CSIP at 0.3.2?
The build log shows 0.3.3 here: https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/159809451#L715

mlubin:
I mean this: https://github.com/SCIP-Interfaces/CSIP/blob/master/src/csip.c#L10

leethargo:
But this is not the latest commit (and the tag is moved, too).

mlubin:
Maybe a github hiccup.

PR #18, opened 2016-09-14
implements getsolvetime

### LINKED ISSUE #17: implement getsolvetime
So I'm starting to run some experiments using SCIP, and I'm going to begin complaining about the missing methods that I need to get things working. First one is ``getsolvetime``.

fserra:
fix #17

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,6 +1,6 @@
 using BinDeps
 
-CSIP_VERSION = "0.3.3"
+CSIP_VERSION = "0.3.4"
 CSIP_URL = "https://github.com/SCIP-Interfaces/CSIP/archive/v$(CSIP_VERSION).zip"
 CSIP_LIB = "libcsip"
 CSIP_UNPACKED = "CSIP-$(CSIP_VERSION)"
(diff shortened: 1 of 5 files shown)


mlubin:
:+1: 
If you've tested it locally let's merge after travis passes.

fserra:
If this is a valid test:
```julia
m = Model(Solver=SCIPSolver())
@variable(m, x)
solve(m)
getsolvetime(m)
```
Then it is good to merge

mlubin:
Yes, it's a valid test (wouldn't hurt to add something to the unit tests).

fserra:
how do you call getsolvetime from mathprogbase??

mlubin:
``MathProgBase.getsolvetime``

fserra:
Ok, this is a mess.

@leethargo we are freeing the transformed problem after the `SCIPsolve` in `CSIPsolve` and so the time obtained by `SCIPgetSolvingTime` is 0.

We have to do something about this,  like a function that needs to be called before modifying a problem.  
Otherwise, we would need to store also the solving time data in CSIP  :anguished: 

How would such a function be integrated? Would we need to call it basically in every method with a ! in its name?

leethargo:
Was problem modification the only issue with the missing `freeTransform` after the `solve`?

fserra:
I think yes

leethargo:
I don't like it, but as with `PySCIPOpt`, it may be unavoidable if we don't want to restrict the feature set.

leethargo:
@fserra : can you add your tests?

fserra:
done

PR #19, opened 2016-09-14
keep a reference to callback userdata

mlubin:
If we don't keep a reference on the Julia side, then the object could be freed by the GC at any time and cause nasty segfaults when we try to access it from inside the callback.

### DIFF
--- a/src/mpb_interface.jl
+++ b/src/mpb_interface.jl
@@ -280,9 +280,9 @@ function setlazycallback!(m::SCIPMathProgModel, f)
     # f is function(d::SCIPLazyCallbackData)
 
     cbfunction = cfunction(lazycb_wrapper, Cint, (Ptr{Void}, Ptr{Void}, Ptr{Void}))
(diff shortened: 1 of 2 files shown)


leethargo:
Is this valid?

CSIP allows multiple callbacks and I know that JuMP is doing some multiplexing on top of MathProgBase, but what if a user adds multiple lazy constraints individually through the MPB interface? I guess that this data would then be overwritten.

But I just saw that the methods in MPB are actually called `set*callback`, so this behaviour is to be expected, I guess.

leethargo:
Anyway, the current implementation of, e.g., `setlazycallback!` in SCIP.jl will just add the new callback in addition to the old one :mag:

mlubin:
Yes, the MPB API expects that at most one callback can be added. Given that this PR fixes the most common use case and doesn't break anything new, I think it's safe to merge.

leethargo:
That's true.

PR #24, opened 2016-09-28
implement solver-independent parameters and fix some compat issues

mlubin:
CC @chriscoey

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.4
 BinDeps
 MathProgBase 0.5 0.6
+Compat 0.8.6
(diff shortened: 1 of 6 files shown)


leethargo:
So there aren't any more solver-independent parameters in MPB right now? I assume that most solvers support limits on nodes, solutions and gap (although the definitions of gap might not agree).

mlubin:
That's all the parameters for now. "I assume" is a lot easier to say than going through all the solvers to make sure that their definitions of these parameters are consistent. We won't add parameters until that work is done by someone who's sufficiently motivated.

mlubin:
@leethargo, safe to tag a new release?

leethargo:
Sure, I remember a discussion of gap and the "Cplex definition" vs the "SCIP definition".

Go ahead with the release, I don't have any changes lined up.

Issue #26, opened 2016-10-24
Support callbacks with nonlinear models


leethargo:
In MathProgBase and JuMP there's a [distinction](http://mathprogbasejl.readthedocs.io/en/latest/solverinterface.html) between solvers for `LinearQuadratic` vs `Nonlinear` (expression-based) problems. According to the [documentation](http://www.juliaopt.org/JuMP.jl/0.14/callbacks.html), the callbacks are supported only for MIP solvers (i.e., `LinearQuadratic`).

For SCIP, this distinction is artificial. We initially implemented the wrapper with a focus on the `LinearQuadratic` interface, later adding support for the `Nonlinear`.

Unfortunately, when building a model using `@NLconstraint`, adding callback functions does not have an effect with SCIP.jl currently (see [example](https://github.com/leethargo/PipeLayout.jl/blob/master/examples/minlp_callbacks.jl)).

@mlubin: Is this a restriction inherent in JuMP or MathProgBase? Can we work around this with little effort?

@fserra: I'm confused by the implementation of `Nonlinear` in SCIP.jl, in particular this code:

```
type SCIPMathProgModel <: AbstractLinearQuadraticModel
[...]
end

type SCIPNonlinearModel <: AbstractNonlinearModel
    m::SCIPMathProgModel
end

LinearQuadraticModel(s::SCIPSolver) = SCIPMathProgModel(s.options)
NonlinearModel(s::SCIPSolver) = SCIPMathProgModel(s.options)
```

The type `SCIPNonlinearModel` is never again used in the code. Might there be a related issue here?

leethargo:
The [MathProgBase documentation](http://mathprogbasejl.readthedocs.io/en/latest/lpqcqp.html#mip-callbacks) gives the signature as `setlazycallback!(m::AbstractLinearQuadraticModel, f)`. So if we actually used `SCIPNonlinearModel` then I would believe this to be the issue. But anyway in `NonlinearModel(s::SCIPSolver)` we return a `SCIPMathProgModel` which is in fact an `AbstractNonlinearModel`?!

mlubin:
> Is this a restriction inherent in JuMP or MathProgBase?

It's not in MathProgBase or JuMP only because we've never put effort into supporting callbacks for nonlinear solvers. I would support tweaking the wording to make it clear that callbacks may be implemented in nonlinear models. JuMP will need a few small changes to set up the callbacks for nonlinear model, shouldn't be too difficult.

[...]

PR #27, opened 2016-10-25
Use proper model types for LinearQuadratic and Nonlinear interfaces

leethargo:
I added two wrapper types and used the more specific method signatures, according to the MathProgBase documentation. For the common methods, I created a union type.

Related to discussion in #26.


mlubin:
Looks fine to me

PR #30, opened 2016-12-09
handle sparse rows in `loadproblem!`

### LINKED ISSUE #4: handle sparse matrix well in `loadproblem`
Only row-wise addition is supported in CSIP.

leethargo:
fixes #4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: julia
 os:
   - linux
 julia:
-  - 0.4
   - 0.5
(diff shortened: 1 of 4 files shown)


leethargo:
Seems to fail on Julia 0.4 because of the difference in how columns of arrays are `1xm` matrices and not vectors :disappointed:

Should we just drop support for 0.4?

mlubin:
Feel free to drop 0.4

PR #31, opened 2016-12-09
rename state :Other -> :Intermediate

leethargo:
as per the change in MathProgBase and JuMP.


mlubin:
@leethargo, just double check that this passes JuMP's tests

leethargo:
Yes, they pass with JuMP/master.

PR #34, opened 2017-01-14
WIP: fix memory leaks

### LINKED ISSUE #33: Memory leaks
There seem to be major memory leaks somewhere between SCIP.jl, CSIP and SCIP.
See this [gist](https://gist.github.com/leethargo/efa15d9fc05b751cf49fe67d8136d969) for a script that realiably fills up all memory on my 4GB notebook and instantly freezes the system :-\

@fserra: do you want to debug this on saturday, and get lunch at Risa afterwards?

leethargo:
fixes #33

### DIFF
--- a/src/mpb_interface.jl
+++ b/src/mpb_interface.jl
@@ -19,8 +19,12 @@ numvar(m::SCIPMathProgModel) = _getNumVars(m)
 numconstr(m::SCIPMathProgModel) = _getNumConss(m)
 
 function freemodel!(m::SCIPMathProgModel)
-    # call finalizer directly
-    freescip(m.inner)
+    if m.inner.ptr_model != C_NULL
+        # call finalizer directly
+        freescip(m.inner)
+    else
+        Base.warn_once("Tried to free already freed model, ignoring.")
+    end
 end
 
 # TODO: mapping for :SemiCont, :SemiInt
--- a/src/mpb_interface.jl
+++ b/src/mpb_interface.jl
@@ -18,7 +18,10 @@ numvar(m::SCIPMathProgModel) = _getNumVars(m)
 "The number of proper constraints, excluding those from lazy callbacks."
 numconstr(m::SCIPMathProgModel) = _getNumConss(m)
 
-freemodel!(m::SCIPMathProgModel) = error("Not implemented for SCIP.jl!")
+function freemodel!(m::SCIPMathProgModel)
+    # call finalizer directly
+    freescip(m.inner)
+end
 
 # TODO: mapping for :SemiCont, :SemiInt
 const vartypemap = Dict{Symbol, Cint}(
--- a/src/types.jl
+++ b/src/types.jl
@@ -13,9 +13,16 @@ type SCIPModel
         # TODO: check return code (everywhere!)
         ccall((:CSIPcreateModel, libcsip), Cint, (Ptr{Ptr{Void}}, ), _arr)
         m = new(_arr[1], options)
-        # QUESTION: Why is _arr not garbage-collected?
+        @assert m.ptr_model != C_NULL
+
+        finalizer(m, freescip)
         return m
     end
+end
+
+function freescip(m::SCIPModel)
+    @assert m.ptr_model != C_NULL
+    _freeModel(m)
 end
 
 # Linear Quadratic Model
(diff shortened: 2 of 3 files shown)


leethargo:
So, SCIP.jl just never called the `freeSCIP`, and SCIP's memory was accumulating. After this is fixed now, there's still memory usage accumulating, even after manual `gc()` calls. But the memory usage is reduced at least partially.

We used SCIP's internal memory debugging tools (`BMSgetMemoryUsed`) and found that, none of the memory tracked by SCIP/BMS is left behind, after a single SCIP instance is created and the Julia wrapper gc'ed. When creating multiple SCIP instances, they share a "memory shell" but are gc'ed individually. Therefore, the reported memory usage is reduced step-by-step, but in the end, completely.

However, according to `top`, not all of the memory is freed properly. Also, `valgrind` found lots of lost/reachable memory:
```
==7748== LEAK SUMMARY:
==7748==    definitely lost: 40 bytes in 1 blocks
==7748==    indirectly lost: 0 bytes in 0 blocks
==7748==      possibly lost: 6,368,351 bytes in 66,227 blocks
==7748==    still reachable: 44,119,928 bytes in 93,188 blocks
==7748==         suppressed: 0 bytes in 0 blocks
==7748== Reachable blocks (those to which a pointer was found) are not shown.
==7748== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==7748== 
==7748== For counts of detected and suppressed errors, rerun with: -v
==7748== Use --track-origins=yes to see where uninitialised values come from
==7748== ERROR SUMMARY: 3976 errors from 2472 contexts (suppressed: 0 from 0)
```

So far, we didn't see any valgrind warnings that point to SCIP.jl code.

leethargo:
@mlubin: how can we go about debugging this properly?

mlubin:
You can use [memory profiling](http://docs.julialang.org/en/stable/manual/profile/#memory-allocation-analysis) to look at the places where Julia allocates memory, this might point in the right direction.

leethargo:
For what it's worth, I tried using the memory profiling. It seems to track only the lines in the top-level source file, so the output is not so helpful in our case (see [gist](https://gist.github.com/leethargo/362e92518830dea05af26db83cf5e669#file-memleaks-jl-mem):
```
        - function leaky(solver, d)
        -     m = Model(solver=solver)
        0     @variable(m, x[1:d] >= 0)
        0     solve(m)
        - end
```

mlubin:
>  It seems to track only the lines in the top-level source file

It creates ``.mem`` files for all julia code used. Look inside ``SCIP.jl/src``.

leethargo:
I feel so stupid now. I suspected that more files would be created, but the git status of SCIP.jl was clean, so I assumed that no extra files are there. Of course, `*.jl.mem` is in `.gitignore` :confused:

leethargo:
I still find it difficult to learn from the `.mem` output.
I ran [another experiment](https://gist.github.com/leethargo/362e92518830dea05af26db83cf5e669#file-memleaks-jl), printing the [output](https://gist.github.com/leethargo/362e92518830dea05af26db83cf5e669#file-gc-log) of SCIP's memory shell.

It shows the same behavior as discussed above: If several SCIP instances are allocated, and some are freed, the SCIP memory shell still shows the memory of the remaining ones. What is interesting, is that the actual calls to the finalizer seems to be somewhat delayed.

If SCIPs are created (and solved and go out of scope?) in a loop, none of them are freed until after the loop is finished and I force `gc()`. Consequently, the memory usage accumulates:
```
run 3x
force garbage collection.
-- calling CSIPfreeModel -- 0 mem used
-- calling CSIPfreeModel -- 19043000 mem used
-- calling CSIPfreeModel -- 9521500 mem used
```

But if I call `gc()` *during* the loop, no memory accumulation takes place:
```
run 3x with gc() after every iteration
-- calling CSIPfreeModel -- 0 mem used
-- calling CSIPfreeModel -- 0 mem used
-- calling CSIPfreeModel -- 0 mem used
force garbage collection.
```

So maybe there is not even a bug, and Julia's garbage collection is just too lazy? Could it be made aware of the indirect memory usage through the C calls, in order to trigger more frequent GCs?

mlubin:
There are no guarantees on when garbage collection will run. Even a call to ``gc()`` is just a suggestion, I think.

> Could it be made aware of the indirect memory usage through the C calls, in order to trigger more frequent GCs?

I'm not aware of a good solution to this, but it's worth posing the question on Discourse.

mlubin:
Note [freemodel!](http://mathprogbasejl.readthedocs.io/en/latest/solverinterface.html#freemodel!) was designed to give users more control over releasing solver resources.

PR #35, opened 2017-02-08
Elaborated installation instruction

Ayush-iitkgp:
Hi,
I had the relatively tough time trying to get SCIP.jl working. This attempt is to make installing SCIP.jl easier for people who don't understand how to set ENV variables and including missing steps during installation. Also I got SCIP.jl working without copying the [patched file](http://scip.zib.de/download/bugfixes/scip-3.2.1/Makefile.doit), I think this step is no longer required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Next set the **environment variable `SCIPOPTDIR`** to point to the directory tha
 
 ```
 export SCIPOPTDIR=/path/to/scipoptdirsuite/sources
-*For example:* export SCIPOPTDIR=/home/hduser/Downloads/scipoptsuite-3.2.1
+Example: export SCIPOPTDIR=/home/hduser/Downloads/scipoptsuite-3.2.1
 ```
 CSIP needs the library in `${SCIPOPTDIR}/lib/scipoptlib.so` and the C header files in `${SCIPOPTDIR}/scip-*/src/`.
 
--- a/README.md
+++ b/README.md
@@ -23,21 +23,22 @@ Julia interface to [SCIP](http://scip.zib.de) solver.
 The SCIP.jl package requires SCIP to be installed.
 
 [Download](http://scip.zib.de/download.php?fname=scipoptsuite-3.2.1.tgz) the
-SCIP Optimization Suite, copy the
-[patched file](http://scip.zib.de/download/bugfixes/scip-3.2.1/Makefile.doit)
-and build the shared library with
+SCIP Optimization Suite and build the shared library with
+
 ```
 make SHARED=true GMP=false READLINE=false ZLIB=false OPT=opt scipoptlib
 ```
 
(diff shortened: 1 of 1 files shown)


leethargo:
Thanks for the effort, I'll have a look later.
1. The failing test is most likely not caused by your changes, I think it has to do with changes in the JuMP tests that we use.
2. I'd be surprised if the patch for SCIP is no longer needed, since they didn't do a release for it. I guess it's not strictly necessary for SCIP.jl to work at all, but it was necessary for something in particular, maybe linking to a `libscipopt.so` that expects IPOPT.

README.md
```diff
@@ -20,23 +20,27 @@ Julia interface to [SCIP](http://scip.zib.de) solver.
 
 ## Installation
 
-The SCIP.jl package requires SCIP to be installed.
+The SCIP.jl package requires [SCIP](http://scip.zib.de/) to be installed.
 
 [Download](http://scip.zib.de/download.php?fname=scipoptsuite-3.2.1.tgz) the
-SCIP Optimization Suite, copy the
-[patched file](http://scip.zib.de/download/bugfixes/scip-3.2.1/Makefile.doit)
```
leethargo: Maybe the story of the patch should be moved somewhere else, together with the reasoning for it (IPOPT support?).

README.md
```diff
@@ -20,23 +20,27 @@ Julia interface to [SCIP](http://scip.zib.de) solver.
...
+```
+CSIP needs the library in `${SCIPOPTDIR}/lib/scipoptlib.so` and the C header files in `${SCIPOPTDIR}/scip-*/src/`.
+
+Clone the SCIP github repository in your local machine and build it in Julia with
 
-Now you should be able to build SCIP.jl in Julia with
 ```
+Pkg.clone("https://github.com/SCIP-Interfaces/SCIP.jl.git")
 Pkg.build("SCIP")
 ```
```
leethargo: Actually, SCIP.jl was registered in `METADATA.jl` since these instructions were written, so could be simplified now.

Ayush-iitkgp: Nice, then we can replace Pkg.build("SCIP") with Pkg.add("SCIP").

README.md
Ayush-iitkgp: Actually, I got an error when I replaced the original Makefile with the above patched file (this is what we were supposed to do.right ?)

I will add your point in the **Notes** below the original instruction.

leethargo: The best (most up-to-date) installation instructions are those that are used in the [Travis build](https://github.com/SCIP-Interfaces/SCIP.jl/blob/master/.travis.yml) :wink:

Ayush-iitkgp: Right :D

Ayush-iitkgp:
@leethargo this attempt is inspired from instructions given in Travis Build file. Tried to state the steps in ordered list with explanation of what each step does.

PR #36, opened 2017-02-17
fixed a bug in function constr_expr_to_nodedata

YankaiCao:


fserra:
Oh nice. We dropped support for 0.4 but didn't check all places properly.

Thanks!

PR #37, opened 2017-02-24
Update bounds for MathProgBase

mlubin:
Enables SCIP to work with JuMP 0.16. Good idea to release a new version following this.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,3 +1,3 @@
-JuMP
+JuMP 0.16
 GLPKMathProgInterface
 FactCheck
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
 julia 0.5
 BinDeps
-MathProgBase 0.5 0.6
+MathProgBase 0.5 0.7
 Compat 0.8.6
--- a/src/mpb_interface.jl
+++ b/src/mpb_interface.jl
@@ -456,12 +456,12 @@ function constr_expr_to_nodedata(ex::Expr)
     values = Float64[]
     csipex = CSIPNodeData[]
 
-    # a ex.args is [side, comp, expr, comp, side] or [expr, comp, side]
+    # a ex.args is [side, comp, expr, comp, side] or [comp, expr, side]
     if length(ex.args) == 5 # two sided constraint
         expr_to_csip(ex.args[3], values, csipex)
     else
         @assert length(ex.args) == 3
-        expr_to_csip(ex.args[1], values, csipex)
+        expr_to_csip(ex.args[2], values, csipex)
     end
     return csipex, values
 end


leethargo:
Yes, but before the release, I'd like to fix some other issues, including the tests that started failing some weeks ago. Not sure what triggered these, maybe some additional tests in JuMP that were introduced, but that never worked with SCIP.jl?

mlubin:
Try again on JuMP 0.16

leethargo:
So you think that the failing tests here are bogus, or resolved in JuMP 0.16?

    Test Summary:                                                                    | Pass  Fail  Error  Total
      Nonlinear                                                                      |   42     1      2     45
      Anonymous nonlinear expression                                                 |          1      1      2

mlubin:
Fixed in JuMP 0.16

leethargo:
In that case, should we put `JuMP 0.16 0.16` into `test/REQUIRE`?

It seems that the tests above do indeed pass, but others, about *MIQP vectorization* fail with 0.16.

leethargo:
Hm, the test result is also different on the latest Travis build, compared to my machine. I hope I'll find some later to actually look into the failing tests.

leethargo:
@fserra: Can you look at this error in the Travis log:

    MethodError: no method matching expr_to_csip(::Symbol, ::Array{Float64,1}, ::Array{SCIP.CSIPNodeData,1})
    Closest candidates are:
      expr_to_csip(!Matched::Expr, ::Any, ::Any) at /home/travis/.julia/v0.5/SCIP/src/mpb_interface.jl:484
      expr_to_csip(!Matched::Number, ::Any, ::Any) at /home/travis/.julia/v0.5/SCIP/src/mpb_interface.jl:507
     in constr_expr_to_nodedata(::Expr) at /home/travis/.julia/v0.5/SCIP/src/mpb_interface.jl:464
     in loadproblem!(::SCIP.SCIPNonlinearModel, ::Int64, ::Int64, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Symbol, ::JuMP.NLPEvaluator) at /home/travis/.julia/v0.5/SCIP/src/mpb_interface.jl:529

Is this maybe a missing case in our handling of nonlinear expressions?

test/REQUIRE
```diff
@@ -1,3 +1,3 @@
-JuMP
+JuMP 0.16
```
mlubin: It's okay but not really needed since this is now the latest release.

leethargo: But in the Travis run after your commit, it actually uses JuMP 0.15.1, as it did on my local machine.

mlubin: Travis ran before 0.16 was released.

fserra:
@leethargo 18511a5 fixes the issue

leethargo:
Wonderful, passing tests :+1: 
It looks like you reverted the changes from #36 again?

fserra:
Yes I reverted #36, I guess I didn't test it and just thought the comment was ok. Though I am curious to why did @YankaiCao got an error in the first place. Maybe because of an older version ?

leethargo:
@mlubin: I think we can do a new release of SCIP.jl now, I don't have any more pending changes.

mlubin:
Cool. You should try out [attobot](https://github.com/attobot/attobot).

PR #39, opened 2017-02-28
add support for conic models

leethargo:


fserra:
Do you why this was/is failing?

leethargo:
I think Amazon's S3 had some issues yesterday that also broke Travis.

fserra:
Oh, that explains why I couldn't look there

leethargo:
Also, I'm not sure if any SOC specific tests exist, so it might just be the same testset as before, I just want to be sure.

PR #44, opened 2017-03-17
WIP: update to new CSIP (and SCIP)

leethargo:


leethargo:
Hm, the build fails on OS X:
```
In file included from src/changesoplex.cpp:19:
src/spxdefines.h:245:1: error: thread-local storage is not supported for the current target
THREADLOCAL extern const Real infinity;
```
It seems to be a new C++ feature that is not (yet) supported by the version of clang shipped by OS X.
The work-around is to choose another compiler, but I don't know how to do that with Travis.

mlubin:
You could possibly [use homebrew](https://docs.travis-ci.com/user/osx-ci-environment/#Homebrew) to install a different compiler version. I'd prefer if the issue with OS X travis didn't hold up releasing the new SCIP.jl, especially since we've never supported OS X in a release yet, so nothing can be broken.

mlubin:
CC @chkwon

chkwon:
Let me see if I can make it work.

chkwon:
[Xcode 8 is required for thread_local.](https://developer.apple.com/library/content/releasenotes/DeveloperTools/RN-Xcode/Introduction.html)

I'll submit a PR soon.

mlubin:
Any objections to merging and tagging this? @chriscoey and I will be running some experiments soon and would like a release of SCIP.jl that works with 4.0.0.

leethargo:
I think we should "unsupport" OS X first, and remove the Travis runs.
Supporting it again can then be a bugfix.

PR #48, opened 2017-03-23
WIP: implement `writeproblem`

leethargo:


leethargo:
I tested with this snippet:
```julia
using SCIP
using JuMP
using MathProgBase.SolverInterface

m = Model(solver=SCIPSolver())
@variable(m, x >= 2)
@constraint(m, 2x <= 7)
@objective(m, :Max, 3x)
solve(m)
im = JuMP.internalmodel(m)
MathProgBase.SolverInterface.writeproblem(im, "foo.lp")
```

But it fails with a segfault (for SCIP compiled in optimized mode). I ran again with SCIP's debug mode, which revealed:
```
julia: src/scip/scip.c:169: checkStage: Assertion `scip->totaltime != NULL' failed.
[...]
checkStage at /home/rs/opt/scipoptsuite-4.0.0/scip-4.0.0/src/scip/scip.c:169
SCIPwriteOrigProblem at /home/rs/opt/scipoptsuite-4.0.0/scip-4.0.0/src/scip/scip.c:10278
writeproblem at /home/rs/.julia/v0.5/SCIP/src/mpb_interface.jl:182
```

I don't really know what this means, maybe the SCIP object is already freed?

fserra:
could it be that you are passing the CSIP pointer instead of the SCIP pointer?

fserra:
yep, it seems that is, I'll push what works for me

leethargo:
Oh, wow, stupid me :astonished:

Issue #28, opened 2016-12-09
Remove CSIP dependency


leethargo:
Now that we have a working version and a better understanding of both Julia and the conceptual gap between SCIP and MathProgBase, we could build a more direct wrapper, skipping CSIP entirely.

We could also (in the same step) provide a Julia wrapper for the "raw" SCIP function(alitie)s, on which the MPB interface implementation is based. This would enable users to easily use the more advanced features directly, that is, without need for `:ccall`. It is unclear to what extent the public functions of SCIP should be made available, since the large number of these suggests a semi-automatic way of generating the wrappers.

CC: @fserra

fserra:
I like this idea a lot.

 How would you like to generate the wrapper? As we have seen, there are two tried approaches. One is using clang.jl and the other is using doxygen and go (i think): https://github.com/ryanjoneil/SCIP.jl

leethargo:
I tried the clang way to generate the first version of the wrapper in SCIP.jl and it was painless for CSIP. But in this case, there were few fancy types involved.

PR #62, opened 2018-01-17
WIP: fix SCIP.jl for julia v0.7

leethargo:


matbesancon:
what's the status on this? I think nightly can be replaced with 1.0 in travis and 0.7 in REQUIRE, this would also help people work with it on 0.7 and allow `Pkg.add("SCIP")` on Julia 1.0

leethargo:
I am not actively working on this right now and am also stuck with 0.6 for my personal work, so this won't change in the near future.

It would make sense to write a new SCIP wrapper for MOI (rather than MathProgBase) and build it diretly on the SCIP C library (rather than the CSIP wrapper, see #28). That could be a whole separate project, then.

matbesancon:
If you're ok I'll try to continue this one. JuMP 0.18.3 now works with Julia 1.0, I think this will be smoother to first upgrade to Julia 1.0 and then to MOI when it is finalized

leethargo:
By all means go for it :+1:

leethargo:
closed in favor of #70

PR #63, opened 2018-01-26
Update to SCIP 5.0.0

leethargo:
SCIP 5.0.0 has significantly improved the performance for MIP and MINLP models.

Right now, one test for MIQP is hitting the timelimit, so the category is skipped for now.


fserra:
I dont  know how to say that I reviewed and agree

leethargo:
I'm not even done yet ;-)
I guess there should be a button on the bottom or on the right-hand side.

PR #70, opened 2018-10-04
v0.7 & v1.0 compatibility

matbesancon:
and couple fixes (using const when possible and when strongly advised)

### DIFF
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,8 +1,6 @@
 using BinDeps
 
-if VERSION >= v"0.7.0-DEV.3382"
-    using Libdl
-end
+using Libdl
 
 include("csip_version.jl")
 
--- a/deps/build.jl
+++ b/deps/build.jl
@@ -1,5 +1,9 @@
 using BinDeps
 
+if VERSION >= v"0.7.0-DEV.3382"
+    using Libdl
+end
+
 include("csip_version.jl")
 
 CSIP_URL = "https://github.com/SCIP-Interfaces/CSIP/archive/v$(CSIP_VERSION).zip"
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,2 +1,2 @@
-JuMP 0.16
+JuMP 0.16 0.19
 OffsetArrays 0.8
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,2 +1,2 @@
 JuMP 0.16
-OffsetArrays 0.2.13
+OffsetArrays 0.8
--- a/test/more_tests.jl
+++ b/test/more_tests.jl
@@ -1,6 +1,6 @@
 # todo: find a meaningful name for this file
 @testset "testing more mpb interface methods" begin
-    m = MathProgBase.LinearQuadraticModel(SCIPSolver("display/verblevel", 0))
+    m = SolverInterface.LinearQuadraticModel(SCIPSolver("display/verblevel", 0))
 
     # take a bit more time integer problem
     A = [314 867 860; 87 875 -695]
@@ -9,15 +9,15 @@
     c = [-1,1,1]
     lb = [40600 -92375]
     ub = lb
-    MathProgBase.loadproblem!(m, A, l, u, c, lb, ub, :Min)
-    @test MathProgBase.numconstr(m) == 2
-    @test MathProgBase.numvar(m) == 3
+    SolverInterface.loadproblem!(m, A, l, u, c, lb, ub, :Min)
+    @test SolverInterface.numconstr(m) == 2
+    @test SolverInterface.numvar(m) == 3
 
-    MathProgBase.setvartype!(m, [:Int, :Int, :Int])
-    MathProgBase.optimize!(m)
-    @test MathProgBase.status(m) == :Optimal
-    @test MathProgBase.getsolvetime(m) >= 0.001
-    @test MathProgBase.getsolution(m) ≈ [750, -200, -25]
+    SolverInterface.setvartype!(m, [:Int, :Int, :Int])
+    SolverInterface.optimize!(m)
+    @test SolverInterface.status(m) == :Optimal
+    @test SolverInterface.getsolvetime(m) >= 0.001
+    @test SolverInterface.getsolution(m) ≈ [750, -200, -25]
 end
 
 modelPath = joinpath(dirname(@__FILE__), "model")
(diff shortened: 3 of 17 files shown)


matbesancon:
cc @leethargo I think travis got stuck because it needs an updated docker script.
More generally, do we need docker for this?

deps/build.jl
```diff
@@ -1,5 +1,9 @@
 using BinDeps
 
+if VERSION >= v"0.7.0-DEV.3382"
```
mlubin: This if condition isn't needed because 0.7 is the minimum supported.

test/REQUIRE
```diff
@@ -1,2 +1,2 @@
 JuMP 0.16
```
mlubin: It's a bad idea to depend on JuMP for testing because all of the existing code will break with the 0.19 release.

matbesancon: Yes that was my conclusion too, but this will require rethinking the tests, I had trouble understanding the mechanisms used: some vectors of solvers are defined before including the JuMP tests.
Maybe in another PR? Freezing the JuMP version to 0.18 upper bound for now

leethargo:
@matbesancon Docker is not really needed for Travis CI. In fact, it was only added recently to avoid compilation of SCIP itself, which was done before a69ecb185c42d49. The SCIP devs now provide a `.deb` package but I did not manage to install that within the Travis script.

test/REQUIRE
leethargo: We could also have a separate test suite, independent from JuMP. There are a couple of tests for CSIP (see [https://github.com/SCIP-Interfaces/CSIP/blob/master/test/test.c](test.c)) that cover a lot of features and would be easily ported, I guess.

test/more_tests.jl
```diff
@@ -1,6 +1,6 @@
 # todo: find a meaningful name for this file
 @testset "testing more mpb interface methods" begin
-    m = MathProgBase.LinearQuadraticModel(SCIPSolver("display/verblevel", 0))
+    m = SolverInterface.LinearQuadraticModel(SCIPSolver("display/verblevel", 0))
```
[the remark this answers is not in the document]
matbesancon: good point I'll see, it was just a reflex to actually call the module where the identifiers are located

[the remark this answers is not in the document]
matbesancon: also, in the REPL:
```julia
import MathProgBase
MathProgBase.<TAB>
```
will not show the re-exported identifiers

matbesancon:
Ok, I think this will be a nice next step to include SCIP directly in the package (with BinaryBuilder + BinaryProvider likely)
In the meantime, I think it requires upgrading your docker image to julia 0.7

leethargo:
> Ok, I think this will be a nice next step to include SCIP directly in the package
I believe that the SCIP license does not allow that. It states that distribution/download needs to be tracked somehow.

> In the meantime, I think it requires upgrading your docker image to julia 0.7
Or, you revert the Travis script to the way it was before, where the SCIP source is downloaded and compiled. Really, it's just a matter of convenience when waiting for CI.

matbesancon:
> Or, you revert the Travis script to the way it was before, where the SCIP source is downloaded and compiled. Really, it's just a matter of convenience when waiting for CI.

The only change in the travis YAML is the julia version, in `travis_docker_test_script.sh` I just made the Pkg syntax 1.0 compatible.

I think the minimal change would be in [the docker script](https://github.com/leethargo/scip-julia-docker/blob/master/files/install-julia.sh) to include 0.7 and 1.0 versions

matbesancon:
> Or, you revert the Travis script to the way it was before, where the SCIP source is downloaded and compiled. Really, it's just a matter of convenience when waiting for CI.

The only change in the travis YAML is the julia version, in `travis_docker_test_script.sh` I just made the Pkg syntax 1.0 compatible.

I think the minimal change would be in [the docker script](https://github.com/leethargo/scip-julia-docker/blob/master/files/install-julia.sh) to include 0.7 and 1.0 versions

matbesancon:
@leethargo I PRed the scip-julia-docker repo, for now this will do, ideally, a next PR could remove the docker dependency all together for travis. Not sure the modification is enough though. Maybe the Dockerfile could also be changed to:
```
RUN apt-get install --no-install-recommends -qq \
  build-essential ca-certificates git rsync unzip wget \
&& /test/install-julia.sh 1.0
```

Depending on DockerHub is a bit of a pain because you cannot work without specifically one person to push a new docker image.

leethargo:
Yes, I think the Dockerfile needs that change, too. But then it will replace the existing Docker image so I can no longer test with Julia 0.6.

I think we can run `install-julia.sh` twice in that Dockerfile, to get both versions. Will try that.

leethargo:
Update: I just pushed a new image to [Docker Hub](https://hub.docker.com/r/leethargo/scip-julia/) which should contain version 1.0 of Julia, while retaining version 0.6, but have not tested it so far.

leethargo:
I triggered the Travis CI run for Julia 1.0 again which now ran successfully :heavy_check_mark:.

@matbesancon: The Travis script also wants to test with Julia 0.7, which is not installed in the Docker image. Should I add this, too, or can it be removed from the CI?

leethargo:
OK, done and done.

After this PR, will the code support *only* versions 0.7 and 1.0?

matbesancon:
yes, the last tagged version with v0.6 will remain available for 0.6 users though

leethargo:
That's fine, but i guess we should update the `REQUIRE` file then, bumping `Julia` to 0.7, and maybe add a line to the README.

Before we tag a new version, we could also try to update the underlying SCIP version from 5 to 6 (I would do that in a separate PR).

mlubin:
FWIW I would recommend tagging a 0.7/1.0 compatible version when it's ready and not blocking it on addition work.

leethargo:
Allright, I somehow missed the `REQUIRE` diff, and will merge then.

As for SCIP 5->6, I think it's "only" performance improvements, so not critical.

matbesancon:
awesome, I might look at the 5->6 upgrade in the coming weeks :)

leethargo:
The update to SCIP 6.0.0 worked without any problem, so I did that quickly before tagging the new version.

However, I just noticed that the tests included from JuMP are not run (anymore). Compare this [old CI run](https://travis-ci.org/SCIP-Interfaces/SCIP.jl/builds/380263911#L874) with the [more recent one](https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/437506807#L604). I remember that we talked about this before, but I must have forgotten about it. It just means that the package is tested only quite superficially in the current version...

mlubin:
> The update to SCIP 6.0.0 worked without any problem, so I did that quickly before tagging the new version.

Awesome :)

> However, I just noticed that the tests included from JuMP are not run (anymore).

That's intended because the tests no longer exist in JuMP 0.19. The right way to do this is to use the corresponding MPB tests and (eventually) MOI tests.

PR #72, opened 2018-10-23
Fix some syntax warnings in Julia 0.7

leethargo:


leethargo:
There is one more warning that I don't understand:

```
┌ Warning: `Array{T}(m::Int) where T` is deprecated, use `Array{T}(undef, m)` instead.
│   caller = setquadobjterms!(::SCIP.SCIPLinearQuadraticModel, ::Array{Int32,1}, ::Array{Int32,1}, ::Array{Float64,1}) at mpb_interface.jl:271
└ @ SCIP ~/.julia/dev/SCIP/src/mpb_interface.jl:271
```

I thought that it might be related to
```
MathProgBase.setquadobj!(m::SCIPLinearQuadraticModel, Q::Array{T, 2}) where {T<:Real} =
    error("Not implemented for SCIP.jl")
```
but changing that does not remove the warning (it's also line 265).

PR #73, opened 2018-10-24
add more tests (in particular from MathProgBase)

leethargo:


leethargo:
It seems that there are some open issues in Coverage with Julia ≥ 0.7, so I will not wait for the report upload to work here.

PR #76, opened 2018-12-18
Rewrite SCIP wrapper with Clang.jl to support a MIP solver through MOI

### LINKED ISSUE #28: Remove CSIP dependency

leethargo:
fixes #28, but does not support callbacks or nonlinear expressions :disappointed:


mlubin:
LQOI wasn't designed to work with nonlinear. What's the motivation for using it instead of directly writing an MOI wrapper? SCIP's API is rich enough that LQOI might not be a good fit (it was designed to wrap APIs that look like Gurobi/CPLEX/Xpress).

@odow @joaquimg

odow:
Hmm yeah LQOI isn't a good fit for SCIP. Best to wrap directly.

leethargo:
Rich _enough_, or _too_ rich? :smirk:

Well, that decision about LQOI was premature, I only looked briefly at the GLPK impementation. What I mean is that at first I will only support SCIP as a MIP solver.

I plan to drop the MPB support, and with it the support for nonlinear constraints and also callbacks. That would be added later. I'm not sure whether it's actually a good idea to do this on the current project, rather than start a new one. But I guess nobody will maintain the old code anyways, and the last version of SCIP.jl to work with JuMP 0.18 is version-tagged, so it should be fine?

cc @fserra

mlubin:
A reason not to drop MPB support yet is that the SCIP package will immediately stop working with JuMP 0.18 and users will have to pin it to an old version to get it working. You'll probably spend more time handling these support questions than you would maintaining the MPB interface for another few months until lots of people are using JuMP 0.19.

mlubin:
OTOH given that it's a complete rewrite I see why you wouldn't want to put the effort into re-enabling MPB support. I didn't consider this in my comment 30 seconds ago :)

leethargo:
Well, yes, I definitely wanted to get rid of the CSIP dependency, so it was a rewrite.
I figure the current release of the code has an upper bound on `MathProgBase` while the new release would have a lower bound on `MathOptInterface`, so the package manage should pick the appropriate version depending on the installed `JuMP` automatically? Maybe that's wishful thinking.

leethargo:
@mlubin Feel free to review and comment. I'm not done yet, but basic LP solving should work now.
In particular, you could suggest additional MOI tests to activate.

I still want to add MIP stuff (variable types, SOS), quadratic constraints and maybe deletion of variables and constraints.

Maybe I'll also expose some dual solution things for LP, but I'm not up-to-date on how easy that is nowadays. In the past, it required going deep into the SCIP-LP-solver interface, but I heard that this has changed.

fserra:
dual lp information is not really possible. we tried to make it work, but
there are fundamental issues, like constraints with one variable get
deleted even if presolve is off. so we have to rely on reduced costs to get
the multipliers, and if there are copies of the same constraint, they will
all get the same multiplier.

src/wrapper/manual_commons.jl
```diff
@@ -0,0 +1,46 @@
+# TODO: created manually, better to improve on wrapper generator!
```
mlubin: Where are these values defined in SCIP? What went wrong with the wrapper generator?

src/SCIP.jl
```diff
@@ -1,39 +1,30 @@
...
-    if csip_installed != csip_required
-        depsdir = realpath(joinpath(dirname(@__FILE__),"..","deps"))
-        error("Installed CSIP version is $(csip_installed), but we require $(csip_required). Run Pkg.build(\"SCIP\") to update.")
+    major = SCIPmajorVersion()
+    minor = SCIPminorVersion()
+    patch = SCIPtechVersion()
+    current = VersionNumber("$major.$minor.$patch")
+    required = VersionNumber("6.0.0")
+    if current < required
+        error("SCIP is installed at version $current, need $required or newer.")
```
mlubin: Is it correct to allow new major versions without regenerating the wrapper?

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
+# holds pointers to SCIP data and takes care of memory mgmt
+mutable struct ManagedSCIP
+    scip::Ref{Ptr{SCIP_}}
```
mlubin: What's the reason to use `Ref{Ptr{SCIP_}}` instead of `Ptr{SCIP_}` directly? (Edit: I see how it's used in `free_scip`. Still, the indirection might be removed for vars and conss.)

src/wrapper.jl
```diff
@@ -0,0 +1,138 @@
...
+include(wrap("cons_sos1"))
+include(wrap("cons_sos2"))
+include(wrap("cons_superindicator"))
+include(wrap("cons_symresack"))
+include(wrap("cons_varbound"))
+include(wrap("cons_xor"))
+
+# SCIP_CALL: macro to check return codes, inspired by @assert
+macro SC(ex)
+    return :(@assert $(esc(ex)) == SCIP_OKAY)
```
mlubin: Users should see error messages instead of assertion failures if something went wrong. Consider creating an `Exception` type that returns the error code and has nice printing. See examples at https://github.com/JuliaOpt/MathOptInterface.jl/blob/master/src/error.jl.

src/wrapper.jl
```diff
@@ -0,0 +1,138 @@
+wrap(base) = joinpath("wrapper", base * ".jl")
+
+# used by Clang.jl
+include(wrap("ctypes"))
+include(wrap("CEnum"))
```
mlubin: Is `CEum` copied from Clang.jl or did you write it? If it's copied then it needs a copyright notice.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+"Returns pointer to SCIP instance"
+get_scip(mscip::ManagedSCIP) = mscip.scip[]
+
+"Returns pointer to SCIP variable"
+get_var(mscip::ManagedSCIP, i::Int) = mscip.vars[i][]
+
+"Returns pointer to SCIP constraint"
+get_cons(mscip::ManagedSCIP, i::Int) = mscip.conss[i][]
+
+"Add variable to problem (continuous, no bounds)"
```
mlubin: Document the return type.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+        mscip.scip[], var, "", -SCIPinfinity(mscip.scip[]),
+        SCIPinfinity(mscip.scip[]), 0.0, SCIP_VARTYPE_CONTINUOUS)
+    @SC rc = SCIPaddVar(mscip.scip[], var[])
+
+    push!(mscip.vars, var)
+    # can't delete variable, so we use the array position as index
+    return length(mscip.vars)
+end
+
+"Add (ranged) linear constraint to problem"
```
mlubin: Document the arguments and the return type.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+function add_variable(mscip::ManagedSCIP)
+    var = Ref{Ptr{SCIP_VAR}}()
+    @SC rc = SCIPcreateVarBasic(
+        mscip.scip[], var, "", -SCIPinfinity(mscip.scip[]),
+        SCIPinfinity(mscip.scip[]), 0.0, SCIP_VARTYPE_CONTINUOUS)
+    @SC rc = SCIPaddVar(mscip.scip[], var[])
+
+    push!(mscip.vars, var)
+    # can't delete variable, so we use the array position as index
+    return length(mscip.vars)
```
mlubin: Consider using a wrapper type like `VariableIndex(length(mscip.vars))` to make it harder to confuse variable indices with constraint indices, as we do in MOI.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+
+"Add variable to problem (continuous, no bounds)"
+function add_variable(mscip::ManagedSCIP)
+    var = Ref{Ptr{SCIP_VAR}}()
+    @SC rc = SCIPcreateVarBasic(
+        mscip.scip[], var, "", -SCIPinfinity(mscip.scip[]),
+        SCIPinfinity(mscip.scip[]), 0.0, SCIP_VARTYPE_CONTINUOUS)
+    @SC rc = SCIPaddVar(mscip.scip[], var[])
+
+    push!(mscip.vars, var)
```
mlubin: Could you `push!(mscip.vars, var[])` and store a vector of pointers instead of a vector of references to pointers?

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+        for var in mscip.vars
+            @SC SCIPreleaseVar(mscip.scip[], var)
+        end
+        @SC SCIPfree(mscip.scip)
+    end
+    @assert mscip.scip[] == C_NULL
+end
+
+"Returns pointer to SCIP instance"
+get_scip(mscip::ManagedSCIP) = mscip.scip[]
```
mlubin: Naming: usually we say `X` instead of `get_X`, e.g., `JuMP.value` instead of `JuMP.get_value`.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
+# holds pointers to SCIP data and takes care of memory mgmt
```
mlubin: Nit: We prefer complete sentences with correct capitalization and periods at the end for comments.

src/managed_scip.jl
```diff
@@ -0,0 +1,74 @@
...
+function free_scip(mscip::ManagedSCIP)
+    # avoid double-free
+    if mscip.scip[] != C_NULL
+        for cons in mscip.conss
+            @SC SCIPreleaseCons(mscip.scip[], cons)
+        end
+        for var in mscip.vars
+            @SC SCIPreleaseVar(mscip.scip[], var)
+        end
+        @SC SCIPfree(mscip.scip)
```
mlubin: Maybe add a note that this sets `mscip.scip` to `C_NULL` because it's not obvious.

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+const LTS = MOI.LessThan{Float64}
+const INS = MOI.Interval{Float64}
+const BOUNDS = Union{EQS, GTS, LTS, INS}
+const BINS = MOI.ZeroOne
+const INTS = MOI.Integer
+const TYPES = Union{BINS, INTS}
+# support changes
+const SCC = MOI.ScalarCoefficientChange{Float64}
+# other MOI types
+const SAT = MOI.ScalarAffineTerm{Float64}
```
mlubin: `SAT` is an overloaded term, potentially a bit confusing to use as an alias. Generally aliases like these should be used in order to make the code more readable and not to save on typing.

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+scip_vartype(::Type{INTS}) = SCIP_VARTYPE_INTEGER
+function MOI.add_constraint(o::Optimizer, func::SVF, set::S) where {S <: TYPES}
+    allow_modification(o)
+    scip = get_scip(o)
+    var = get_var(o, func.variable)
+    infeasible = Ref{Ptr{SCIP_Bool}}
+    @SC SCIPchgVarType(scip, var, scip_vartype(S), infeasible[])
+    if S <: BINS
+        # need to adjust bounds for SCIP?!
+        @SC SCIPchgVarLb(scip, var, 0.0)
```
mlubin: How does this interact with variable bounds? I.e., what if you have `x == 1, Binary`. Does this drop the tighter bounds? I'd also want an error/infeasibility report for `2  <= x <= 3, Binary`.

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+    return register!(o, CI{SVF, S}(i))
+end
+
+function MOI.add_constraint(o::Optimizer, func::SVF, set::S) where S <: BOUNDS
+    allow_modification(o)
+    var = get_var(o, func.variable)
+    lb, ub = bounds(set)
+    lb == nothing || @SC SCIPchgVarLb(get_scip(o), var, lb)
+    ub == nothing || @SC SCIPchgVarUb(get_scip(o), var, ub)
+    # use var index for cons index of this type
```
mlubin: This is the right index to use (https://github.com/JuliaOpt/MathOptInterface.jl/issues/570). Is a useful error returned if a user tries to add bounds on the same variable twice?

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+    rhs = rhs == nothing ?  SCIPinfinity(scip) : rhs
+
+    @SC SCIPchgLhsLinear(scip, cons, lhs)
+    @SC SCIPchgRhsLinear(scip, cons, rhs)
+
+    return nothing
+end
+
+function MOI.is_valid(o::Optimizer, ci::CI{SAF,<:BOUNDS})
+    1 <= ci.value <= length(o.mscip.cons)
```
mlubin: Nit: Use explicit `return` except for one-line functions (https://github.com/JuliaOpt/JuMP.jl/issues/1718).

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+    end
+    constant = SCIPgetOrigObjoffset(get_scip(o))
+    return SAF(terms, constant)
+end
+
+function MOI.set(o::Optimizer, ::MOI.ObjectiveSense, sense::MOI.OptimizationSense)
+    allow_modification(o)
+    if sense == MOI.MIN_SENSE
+        @SC SCIPsetObjsense(get_scip(o), SCIP_OBJSENSE_MINIMIZE)
+    elseif sense == MOI.MAX_SENSE
```
mlubin: `FEASIBILITY_SENSE` should also be handled, or at least note that nothing needs to be done.

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,414 @@
...
+
+term_status_map = Dict(
+    SCIP_STATUS_UNKNOWN => MOI.OPTIMIZE_NOT_CALLED,
+    SCIP_STATUS_USERINTERRUPT => MOI.INTERRUPTED,
+    SCIP_STATUS_NODELIMIT => MOI.NODE_LIMIT,
+    SCIP_STATUS_TOTALNODELIMIT => MOI.NODE_LIMIT,
+    SCIP_STATUS_STALLNODELIMIT => MOI.OTHER_LIMIT,
+    SCIP_STATUS_TIMELIMIT => MOI.TIME_LIMIT,
+    SCIP_STATUS_MEMLIMIT => MOI.MEMORY_LIMIT,
+    SCIP_STATUS_GAPLIMIT => MOI.OTHER_LIMIT,
```
mlubin: Other solvers map this to `OPTIMAL`. I.e., terminating with a gap is optimal within the user-provided termination criteria.

test/REQUIRE
```diff
@@ -1 +1 @@
-JuMP
+
```
mlubin: This file can be removed if it's empty.

test/managed_scip.jl
```diff
@@ -0,0 +1,45 @@
...
+
+@testset "create and implicit free" begin
+    ptr = Ref{Ptr{SCIP.SCIP_}}(C_NULL)
+    for i = 1:1
+        mscip = SCIP.ManagedSCIP()
+        ptr = mscip.scip
+        @test ptr[] != C_NULL
+    end
+    GC.gc()
+    GC.gc() # TODO: why need a second call?
```
mlubin: I think there's no guarantee when `finalize` runs. It's better just test that explicit `finalize` works. Depending on the behavior of `GC.gc()` could break the test in different julia versions.

REQUIRE
```diff
@@ -1,3 +1,3 @@
-julia 0.7
-BinDeps
-MathProgBase 0.5 0.8
+julia 1.0
+Compat
```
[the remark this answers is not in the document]
leethargo: This might be true. It is used in [build.jl](https://github.com/SCIP-Interfaces/SCIP.jl/blob/7e4054b185b360ff428e0de1f7985906137ee57e/deps/build.jl#L20) but that was just copy-pasted from some other solver wrapper.

src/wrapper.jl
leethargo: Yes, this is kind of hack. I was playing with a larger macro that would also show the file and line number, but I could not get it to work quickly.

Note: with the old SCIP.jl we were simply ignoring these return codes.

src/wrapper/README.md
```diff
@@ -0,0 +1,2 @@
+These files are automatically generated with `Clang.jl`. See
+`gen/generate_wrapper.jl` for details.
```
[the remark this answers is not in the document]
leethargo: I will open a separate issue for that. I had several problems with the tagged version of Clang.jl, but there was a rewrite of it that I used from the PR branch directly. I think was since merged to master.

Also, the generation is not fully automatic (yet), because some `#define`s were not wrapped (they appear as comments in the generated files), I had to rename `Ptr{SCIP}` to `Ptr{SCIP_}` to avoid a name conflict with the module and finally deleted a bunch of mostly empty files (e.g. for `type_var.jl`).

src/wrapper/manual_commons.jl
leethargo: These are defined in the SCIP headers that I wrap, but for some reason (did not investigate yet) were skipped by Clang.jl.

The `FILE` type is just standard C, I guess? It's not defined by SCIP itself, at least.

src/SCIP.jl
leethargo: OK, added another check against an upper bound.

src/managed_scip.jl
leethargo: Done (for existing docstrings).

src/managed_scip.jl
leethargo: Actually no, we do need `SCIP_VAR**` for `SCIPreleaseVar`, so I don't think we can do anything simpler.

See also [discussion on discourse](https://discourse.julialang.org/t/confusion-about-ptr-ref-and-dereferencing/18805).

src/managed_scip.jl
leethargo: Hm, yes. I wondered that if I added a function `var()`, it might conflict with local variables of the same name?

src/managed_scip.jl
leethargo: No (see above).

src/wrapper.jl
leethargo: Are you suggesting to mention these files in the `LICENSE` file? If we add a note in the header, these would be overwritten by the wrapper generation script.

src/wrapper.jl
leethargo: I want to handle this in a separate issue: #78.

src/wrapper/README.md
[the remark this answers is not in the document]
leethargo: to be handled separately: #77

src/wrapper/manual_commons.jl
leethargo: see #77

src/managed_scip.jl
leethargo: Am using `scip`, `var` and `cons` now.

src/managed_scip.jl
leethargo: OK, I added types `VarRef` and `ConsRef` to avoid confusion with the types from MOI.

src/wrapper.jl
leethargo: I added a line to each of the files.

src/MOI_wrapper.jl
leethargo: Well, neither of these work right now.

I found that SCIP fails with an error, if I change the variable type to binary but the bounds are not `[0, 1]`.

To support the single variable constraints properly, we would have to store copies of all of them, I guess, which is something I wanted to avoid. Alternatively, we could create linear constraints for them, and leave the variable bounds at their defaults (-/+ infinity).

src/MOI_wrapper.jl
leethargo: OK, many of these aliases were only used on one or two places, so I removed several of them.

src/MOI_wrapper.jl
leethargo: Hm, we could reset the objective coefficients, I guess. But then they could not be recovered easily.

src/MOI_wrapper.jl
leethargo: Not yet, see [MOI/#570](https://github.com/JuliaOpt/MathOptInterface.jl/issues/570#issuecomment-451742785)

src/MOI_wrapper.jl
```diff
@@ -0,0 +1,399 @@
...
+const MOIU = MOI.Utilities
+
+# indices
+const VI = MOI.VariableIndex
+const CI = MOI.ConstraintIndex
+# supported functions
+const SVF = MOI.SingleVariable
+const SAF = MOI.ScalarAffineFunction{Float64}
+# supported sets
+const Bounds = Union{MOI.EqualTo{Float64}, MOI.GreaterThan{Float64},
```
mlubin: Nit: Constants should be all upper-case (http://www.juliaopt.org/JuMP.jl/dev/style/#Naming-1), e.g., `BOUNDS` and `VAR_TYPES`.

src/MOI_wrapper.jl
mlubin: Consider adding a TODO to catch this case and make it an error.

mlubin:
(Unable to reply inline)

> Other solvers map this to OPTIMAL. I.e., terminating with a gap is optimal within the user-provided termination criteria.

> Hm, strange. But what if the solver "accidentally" proves optimality (zero gap)? These cases could not be distinguished, then, using the status?

Correct. If you set a convergence tolerance on the solver and the solver happens to do better than you asked, you still get OPTIMAL. You can always check the gap manually later.

src/MOI_wrapper.jl
leethargo: See discussion in #79

src/MOI_wrapper.jl
leethargo: Added a warning, and opened issue #80.

src/MOI_wrapper.jl
leethargo: OK, done

src/MOI_wrapper.jl
leethargo: Yeah, I felt like this applies more to numerical constants etc., but I don't care enough myself and wish we had a working formatting tool.

PR #81, opened 2019-01-06
Deal with implicit and duplicate constraints for variable bounds

### LINKED ISSUE #79: Binary variables with bounds
How does this interact with variable bounds? I.e., what if you have `x == 1, Binary`. Does this drop the tighter bounds? I'd also want an error/infeasibility report for `2  <= x <= 3, Binary`.

_Originally posted by @mlubin in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MTg4NTYwNjMz/pull_request_reviews/more_threads_

I couldn't reply to the comment on #76 for some reason, so opening an issue. The current code in #76 seems to silently ignore user-provided bounds on binary variables.

leethargo:
fixes #79 

Still needs additional tests. Maybe prints too many warnings now. Probably is too conservative (does not consider variable type when throwing error about existing bounds).

In particular, will still fail for these cases now:
- fixing a binary variable
- having two separate constraints, one for the lower bound, one for the upper bound

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -204,8 +204,8 @@ function MOI.set(o::SCIP.Optimizer, ::MOI.ConstraintSet, ci::CI{SVF,S}, set::S) 
     allow_modification(o)
     v = var(o, VI(ci.value)) # cons index is actually var index
     lb, ub = bounds(set)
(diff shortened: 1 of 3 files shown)


leethargo:
Could still add some tests (and handling) of changes in the constraint set for var bounds.

PR #82, opened 2019-01-09
Support quadratic constraints

leethargo:


leethargo:
Hm, my implementation seems mature now (some unpushed commits), but the QCP tests of MOI fail, because SCIP does not find the correct soluion?!

Will report bug with SCIP team...

PR #88, opened 2019-01-14
WIP: support deletion of variables and constraints

### LINKED ISSUE #85: Support MOI.delete for variables and constraints
It should not be so difficult for SCIP itself, but requires a rethinking of how we decide on variable and constraint indices, as well as additional mapping structures that might become necessary.

leethargo:
fixes #85


leethargo:
Problem: I use `SCIPdelVar` to remove the variable from SCIP (after calling `SCIPfreeTransform`, so we are in `STAGE_PROBLEM`) which in turn calls `SCIPprobDelVar`, but this does not delete variable from the constraints it is part of, which leads to problems later.

```
/** marks variable to be removed from the problem; however, the variable is NOT removed from the constraints */
extern
SCIP_RETCODE SCIPprobDelVar(
   SCIP_PROB*            prob,               /**< problem data */
   BMS_BLKMEM*           blkmem,             /**< block memory */
   SCIP_SET*             set,                /**< global SCIP settings */
   SCIP_EVENTQUEUE*      eventqueue,         /**< event queue */
   SCIP_VAR*             var,                /**< problem variable */
   SCIP_Bool*            deleted             /**< pointer to store whether marking variable to be deleted was successful */
   );
```
```
julia: /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/cons_linear.c:1025: consdataCreate: Assertion `(*consdata)->vars[v] != NULL' failed.

signal (6): Aborted
in expression starting at /home/rs/.julia/dev/SCIP/test/MOI_wrapper.jl:10
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7fc4b45ede66)
__assert_fail at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
consdataCreate at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/cons_linear.c:1025
consTransLinear at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/cons_linear.c:15664
SCIPconsTransform at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/cons.c:6427
SCIPprobTransform at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/prob.c:575
SCIPtransformProb at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/scip_solve.c:450
SCIPpresolve at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/scip_solve.c:2440
SCIPsolve at /home/rs/opt/scipoptsuite-6.0.0/scip/src/scip/scip_solve.c:2660
SCIPsolve at /home/rs/.julia/dev/SCIP/src/wrapper/scip_solve.jl:14
unknown function (ip: 0x7fc49c0b9c55)
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2184
optimize! at /home/rs/.julia/dev/SCIP/src/MOI_wrapper.jl:137
linear5test at /home/rs/.julia/packages/MathOptInterface/g50PO/src/Test/contlinear.jl:700
```

leethargo:
@fserra, any idea on how to identify constraints (by `SCIP_CONS*`) that involve a specific `SCIP_VAR*`, short of storing the whole variable-constraint incidence matrix?

I have found `SCIPgetConsVars`, which is generic, but brute force :cry:

fserra:
I actually don't know. I can imagine that this functionality does really make much sense in SCIP given that it is constraint-based and does not really have a matrix.

So I am guessing here that deleting basically means deleting from a linear problem?

I would rename the variable and set the bounds to 0.

For more general constraints I wouldn't know what to do... what does deleting `x` means in `y/x`?

fserra:
Oh I was pretty sure I wrote some more stuff but apparently it wasn't the case.

So I had a look at `SCIPdelVar` and it actually calls `SCIPprobPerformVarDeletions` after calling `SCIPprobDelVar`.

The function  `SCIPprobPerformVarDeletions` goes through the constraint handlers and they delete the variable.

leethargo:
I saw the `SCIPprobPerformVarDeletions`, too, but I thought it would only remove it from auxiliary data structures, such as clique table etc. I also figured that this deletion only makes sense for some constraint types, such as linear. In this case, the assertion is in `cons_linear`, actually.

I will try to recreate the problem with minimal C code.

leethargo:
Maybe [this question](https://scip.zib.de/doc-6.0.1/html/FAQ.php#howtodeletevariables) from the FAQ is also relevant:

> SCIP features the functionality to delete variables from the problem when performing branch-and-price. This feature is still in a beta status and can be activated by switching the parameters pricing/delvars and pricing/delvarsroot to TRUE in order to allow deletion of variables at the root node and at all other nodes, respectively. Furthermore, variables have to be marked to be deletable by SCIPvarMarkDeletable(), which has to be done before adding the variable to the problem. Then, after a node of the branch-and-bound-tree is processed, SCIP automatically deletes variables from the problem that were created at the current node and whose corresponding columns were already removed from the LP. Note that due to the way SCIP stores basis information, it is not possible to completely delete a variable that was created at another node than the current node. You might want to change the parameters lp/colagelimit, lp/cleanupcols, and lp/cleanupcolsroot, which have an impact on when and how fast columns are removed from the LP.

> Constraint handlers support a new callback function that deletes variables from constraints in which they were marked to be deleted. Thus, when using automatic variable deletion, you should make sure that all used constraint handlers implement this callback. By now, the linear, the set partitioning/packing/covering and the knapsack constraint handler support this callback, which should be sufficient for most branch-and-price applications. Note that set covering constraints can be used instead of logicor constraints.

> Instead of deleting a variable completely, you can also remove it from the problem by either fixing the variable to zero using SCIPfixVar(), which fixes the variable globally or using SCIPchgVarUbNode() and SCIPchgVarLbNode(), which changes the bounds only for the current subtree.

So, `cons_linear` should support in, theoretically, but most other constraint handlers probably won't.

fserra:
I'll have a look

Yeah, but that documentation is for branch and price, here you are just in PROBLEM stage and everything should just work there.

fserra:
ok, this is tricky.

The problem is that in your program the constraint is not added to the constraint handler so the constraint handler doesn't have a chance to go through its constraints and delete the var.

Furthermore, to make deletion more efficient, constraints are marked as having a variable to be deleted using the event handler. However, the event handlers are not installed before solving starts.

so to make this work would require a bit of work on the scip side... it would only be available the next release

leethargo:
Thanks for the research!

> The problem is that in your program the constraint is not added to the constraint handler so the constraint handler doesn't have a chance to go through its constraints and delete the var.

Why? Should I call some other method to make the constraint handler aware of the constraint?

> However, the event handlers are not installed before solving starts.

OK, so the `SCIPdelVar` should get another stage check (and this is a bug)?
Would it help to manually start the event handlers?

For example, I will already call `SCIPfreeTransform` from SCIP.jl everytime I want to modify the problem, in order to get back to problem stage. I could maybe do something similar to initialize the solving process? Not sure if the event handlers are killed with `SCIPfreeTransform`?

> so to make this work would require a bit of work on the scip side... it would only be available the next release

And even then it would only work for "supported constraint handlers" (implementing the callback), right?

I guess for now, the best would be to only support deletion of constraints through SCIP.jl, and throw an error when a variable is deleted while constraints still exist.

fserra:
> Thanks for the research!
> 
> > The problem is that in your program the constraint is not added to the constraint handler so the constraint handler doesn't have a chance to go through its constraints and delete the var.
> 
> Why? Should I call some other method to make the constraint handler aware of the constraint?

I don't think so. From what I saw in the code, the constraint is added to the conss array that SCIP knows, but only when transforming you add the constraints to the conshldrs conss array. I guess it makes sense, why would the conshldr like to know the constraints in the original problem?

> > However, the event handlers are not installed before solving starts.
> 
> OK, so the `SCIPdelVar` should get another stage check (and this is a bug)?

Well, it is definitely a bug at least from the function names and documentation. Basically, you cannot really call this function in problem stage.

> Would it help to manually start the event handlers?

I don't think you can manually start an event handler. This is an event handler that the linear constraint handler creates. However, it starts catching events in CONSINIT and this happens after transforming the problem.
 
> For example, I will already call `SCIPfreeTransform` from SCIP.jl everytime I want to modify the problem, in order to get back to problem stage. I could maybe do something similar to initialize the solving process? Not sure if the event handlers are killed with `SCIPfreeTransform`?

In CONSEXIT, which is called before the transformed problem is freed, the events are dropped, so it still wouldn't work

> > so to make this work would require a bit of work on the scip side... it would only be available the next release
> 
> And even then it would only work for "supported constraint handlers" (implementing the callback), right?

Yes

> I guess for now, the best would be to only support deletion of constraints through SCIP.jl, and throw an error when a variable is deleted while constraints still exist.

yeah... well the idea of scip is mostly (modulo reoptimization) to do everything within scip and not to solve an then solve again, so all this 'features' are shaky

PR #100, opened 2019-02-22
Support nonlinear constraints.

### LINKED ISSUE #93: Support constraints with general nonlinear expression.
Since there have been no changes from MPB to MOI (?) this might be comparatively easy, since we could look at CSIP for inspiration.

But if that interface is subject to change in MOI, we might also wait and do it later.

leethargo:
Fixes #93.

This adds support for nonlinear constraints based on `:ExprGraph`. The Julia expression is converted to an intermediate array-based representation which is then converted to a SCIP-style expression graph.

The code is mostly translated from CSIP and the old MathProgBase wrapper. However, the handling of powers (with constant exponent) and unary minus is done earlier, to avoid memory leaks or worse.

I hope that the tests that I created use the same expression format that would also be generated by JuMP. I'm tempted to use JuMP for additional tests, but don't want cyclic dependencies.

EDIT: No longer using the intermediate representation, so the CSIP layer is basically removed now.
EDIT: Still waiting for input on how to fix the non-deterministic behavior related to constant expressions.


leethargo:
Hm, the Travis tests are inconsistent. Locally, the results seem to be non-deterministic:
- most of the time, all tests pass
- sometimes, the first nonlinear test fails (status: unbounded)
- sometimes, the second nonlinear test fails (status: infeasible)

src/MOI_wrapper/nonlinear_constraints.jl
```diff
@@ -0,0 +1,124 @@
+# Nonlinear constraints (objective not supported)
+
+MOI.supports(::Optimizer, ::MOI.NLPBlock) = true
+
+function MOI.set(o::Optimizer, ::MOI.NLPBlock, data::MOI.NLPBlockData)
+    # We don't actually store the data (to be queried later during the solve
+    # process). Instead, we extract the expression graphs and add the
+    # corresponding constraints to the model directly.
+    if data.has_objective
+        @warn "Nonlinear objective not supported by SCIP.jl!" maxlog=1
```
mlubin: Shouldn't this be an error instead of a warning if the solver isn't capable of solving the problem that the user wrote down?

src/MOI_wrapper/nonlinear_constraints.jl
```diff
@@ -0,0 +1,124 @@
...
+    :* => SCIP_EXPR_PRODUCT,
+    :- => SCIP_EXPR_MINUS,
+    :/ => SCIP_EXPR_DIV,
+    :^ => SCIP_EXPR_REALPOWER,
+    :sqrt => SCIP_EXPR_SQRT,
+    :exp => SCIP_EXPR_EXP,
+    :log => SCIP_EXPR_LOG,
+)
+
+mutable struct NonlinExpr
```
mlubin: Could you add a few comments on this data structure?

leethargo: Good idea, especially with the abundance of special cases in the translation.

src/MOI_wrapper/nonlinear_constraints.jl
leethargo: True. But there is also no way for the solver to signal that it can't handle them, because the `supports` method is just about `NLPBlock` (all or nothing).

mlubin: `error()` is a valid way to signal this :). I think that's better than returning a solution that is almost surely incorrect. Warnings are helpful but it's possible to miss them.

leethargo:
Output for the unbounded result:
```
feasible solution found by trivial heuristic after 0.0 seconds, objective value 0.000000e+00
presolving:
(round 1, fast)       0 del vars, 0 del conss, 0 add conss, 0 chg bounds, 0 chg sides, 0 chg coeffs, 1 upgd conss, 0 impls, 0 clqs
problem infeasible or unbounded: variable <t_> with objective -1 can be made infinitely large
presolving (2 rounds: 2 fast, 0 medium, 0 exhaustive):
 1 deleted vars, 0 deleted constraints, 0 added constraints, 0 tightened bounds, 0 added holes, 0 changed sides, 0 changed coefficients
 0 implications, 0 cliques
transformed 1/1 original solutions to the transformed problem space
presolving detected unboundedness
Presolving Time: 0.00

SCIP Status        : problem is solved [unbounded]
Solving Time (sec) : 0.00
Solving Nodes      : 0
Primal Bound       : -1.00000000000000e+20 (1 solutions)
Dual Bound         : -1.00000000000000e+20
Gap                : 0.00 %
```

odow:
Let me know if you need a hand interfacing MINLPTests. It is definitely the way to go. You should be able to pass all nlp and nlp-cvx tests. There are some issues with the mixed integed ones.

leethargo:
OK, so I feel there is still some work on the table, in several steps:

- [ ] identify the source of the non-deterministic behavior in tests (current theory: a Julia array that is passed to SCIP in a ccall is GC'd before SCIP has created a copy, maybe [here](https://github.com/SCIP-Interfaces/SCIP.jl/pull/100/files#diff-9582432650e4a813bffacffcc37711c6R292) or [here](https://github.com/SCIP-Interfaces/SCIP.jl/pull/100/files#diff-9582432650e4a813bffacffcc37711c6R300)).
- [ ] remove the intermediate array-based representation of the expression graph which was mostly useful for the CSIP layer in C.
- [ ] add more comprehensive tests through JuMP with MINLPTests.

leethargo:
Unfortunately, adding a `GC.preserve` block with the arrays used around the SCIP calls did not change the behavior (see [gist](https://gist.github.com/leethargo/1cd7d8f56af07c74570bd82d12221f03).)

leethargo:
While having SCIP print the problem in its representation, I found that weird values are shown for the constants in the failing cases (rather than `2.0`):

```
CONSTRAINTS
  [nonlinear] <c0>: ((4.94066e-324 * <x0>) + <x1>) <= 1.5;
END
```
```
CONSTRAINTS
  [nonlinear] <c0>: ((6.90208e-310 * <x0>) + <x1>) <= 1.5;
END
```

These look like access to unitialized memory. I checked the values just before the `ccall` where it was still OK. Also the value is passed as `Cdouble`, not via a `Ptr` or in an array.

Just to be sure, I added `GC.enable(false)` and `GC.enable(true)` around all the includes in `run_tests.jl` which did not make any difference, so I'm ruling that out for now...

leethargo:
In the other testset, it becomes more apparent that *all* constant values (inside the expressions) are corrupted:
```
CONSTRAINTS
  [nonlinear] <c0>: (<x0> + 4.94066e-324) == 2;
  [nonlinear] <c1>: (<x1> - <x2>) == 2;
  [nonlinear] <c2>: ((4.94066e-324 - <x3>) + 6.9532e-310) == 2;
  [nonlinear] <c3>: (<x4> + <x5> + <x6>) == 2;
  [nonlinear] <c4>: (<x7> * <x8> * <x9>) == 2;
  [nonlinear] <c5>: realpower((<x10> + <x11>), 6.91388e-310) == 2;
  [nonlinear] <c6>: (<x12> / <x13>) == 2;
  [nonlinear] <c7>: sqrt(<x14>) == 2;
  [nonlinear] <c8>: exp(<x15>) == 2;
  [nonlinear] <c9>: log(<x16>) == 2;
END
```
In particular, the second argument of `realpower` is affected, even though it is not a proper sub-expression of type `SCIP_EXPR_CONST`.

leethargo:
I also found that if I compile with cmake option `-DNOBLKBUFMEM=ON`, I can not reproduce the failing behavior.

@fserra, do you have an idea what this would imply?
I also ran valgrind on that test, but it does not look like it found anything relevant ([gist](https://gist.github.com/leethargo/d5e315ca9fc95a545d86ec02e2ea8141)).

leethargo:
Out of desperation, I also reimplemented the test in C, calling SCIP directly ([gist](https://gist.github.com/leethargo/f5781e1fe17adb0a89d9b5b13a049de7)). In this case, it always seems to work. Valgrind does not see problems either.

src/MOI_wrapper/nonlinear_constraints.jl
leethargo: I refactored to code instead, to not use this data structure anymore.

leethargo:
The refactoring was not related to it, but the code is still non-deterministic, w.r.t. to memory corruption.
Interestingly, the output is different now, in that the `Cdouble` argument of `REALPOWER` is affected anymore, only the numbers in the `CONST` expressions:
```
CONSTRAINTS
  [nonlinear] <c0>: (<x0> + 6.93687e-310) == 2;
  [nonlinear] <c1>: (<x1> - <x2>) == 2;
  [nonlinear] <c2>: ((6.9369e-310 - <x3>) + 1) == 2;
  [nonlinear] <c3>: (<x4> + <x5> + <x6>) == 2;
  [nonlinear] <c4>: (<x7> * <x8> * <x9>) == 2;
  [nonlinear] <c5>: realpower((<x10> + <x11>), 0.8) == 2;
  [nonlinear] <c6>: (<x12> / <x13>) == 2;
  [nonlinear] <c7>: sqrt(<x14>) == 2;
  [nonlinear] <c8>: exp(<x15>) == 2;
  [nonlinear] <c9>: log(<x16>) == 2;
END
```

leethargo:
I was able to narrow down the problem further:
After creating the SCIP expression for constant values, that value is sometimes corrupted directly afterwards. That is, after the call to [SCIPexprCreate](https://github.com/SCIP-Interfaces/SCIP.jl/pull/100/files#diff-098582d81e432d120cd9aab2b5f231e7R125), I would get
```
value = 2.0
expr__ = Base.RefValue{Ptr{Nothing}}(Ptr{Nothing} @0x00000000027f7050)
SCIPexprGetOpReal(expr__[]) = 6.89937396681325e-310
SCIPexprGetOpIndex(expr__[]) = -1682932864
SCIPexprGetOpData(expr__[]) = Ptr{Nothing} @0x00007f019bb07b80
```

The three methods `SCIPexprGet*` all get the "same" value, which is a C `union`.
In `SCIPexprCreate`, that union `opdata` is actually stack allocated, but then assigned to a `struct` member of `SCIP_EXPR*`, so that shouldn't be a problem.

fserra:
does it also go away if you just disable block memory (-DNOBLKMEM)? The idea of block memory is that scip allocates memory itself and the re-uses it. I don't see how this can cause this though

leethargo:
Yes, adding `-DNOBLKMEM` is enough to avoid the problem.

I tried to create minimal examples for both the C code and (equivalent) Julia code that are both self-contained: [gist](https://gist.github.com/leethargo/d55a14594213f48fdd5e4c266c09bf05).

fserra:
I guess the problem has to be the way julia handles variadic c functions. I implemented a `SCIPexprCreate22` which receives a `SCIP_Real` instead of `...` and only creates constant expressions.
Then I modified the julia script so that it calls that function instead and after running 400 times the test it never failed.

mlubin:
The Julia signatures for the call to `SCIPexprCreate` don't seem to match what the documentation says to do for variadic functions: https://docs.julialang.org/en/v1/manual/calling-c-and-fortran-code/#man-bits-types-1. There should be `...` somewhere. (I haven't done this myself.)

leethargo:
Yes, that must be it. Changing the signature for `ccall` does seem to resolve issue for me.
Unfortunately, several operators (`SUM`, `PRODUCT`, `REALPOWER`) require variadic arguments of different types, so I guess we can't get around introducing another layer of C inbetween.

test/MINLPTests/run_minlptests.jl
```diff
@@ -0,0 +1,24 @@
...
+const DUALTOL = NaN  # to disable the query
+
+@testset "MINLPTests" begin
+    @testset "nlp_005_010" begin
+        MINLPTests.nlp_005_010(OPTIMIZER, OBJTOL, PRIMALTOL, DUALTOL;
+                               termination_target = MOI.OPTIMAL)
+    end
+
+    @testset "nlp_005_011" begin
+        MINLPTests.nlp_005_010(OPTIMIZER, OBJTOL, PRIMALTOL, DUALTOL;
```
odow: It looks like this should be `nlp_005_011`?

leethargo: Yes, thanks. There are also more to come from the `nlp_cvx` set...

leethargo:
> Any reason to only test a couple of MINLPTests instead of the full set (excluding the ones with trig functions etc)?

Yes: I had to change the `target_termination` to `OPTIMAL` or `INFEASIBLE`, rather than the `LOCAL_` variants.

Also, I handpicked the few that are supported, which is a smaller list than the excluded ones:
- SCIP itself [does not support trigonometric functions](https://scip.zib.de/doc-6.0.1/html/cons__nonlinear_8c_source.php#l3115).
- SCIP.jl only supports affine objective functions ( I guess it can also do `Min -x` but not `Min x`). So with a suitable objective bridge in MOI, I could add many more.

odow:
> So with a suitable objective bridge in MOI

It's not very had to add support for `SingleVariable` objective functions as a temporary stop-gap? It can be removed once there is a bridge.

leethargo:
> It's not very had to add support for SingleVariable objective functions as a temporary stop-gap? It can be removed once there is a bridge.

Well, I meant that with an objective bridge, we could support quadratic and nonlinear objectives. The objective bridge itself will probably require SingleVariabe objectives, so that it makes sense to support it manually.

test/MINLPTests/run_minlptests.jl
leethargo: I now added all tests that I think can be supported now. For one test, SCIP had a wrong solution, which I will try to investigate upstream.

leethargo:
So, the major thing left to do is the C wrapper for `SCIPexprCreate`.

Actually, I'm not so sure how to do that after all. For `SCIP_EXPR_SUM`, I will need to call `SCIPexprCreate` with an unknown number of arguments. The first one will be `int nchildren`, but then there is a set of `SCIP_EXPR*`. It's not possible to simply [forward variadic arguments](http://c-faq.com/varargs/handoff.html) in C function calls, without providing another version of the called function (with a different signature).

The only workaround I can figure out now is to pass an array of `SCIP_EXPR*` and then prepare many different calls, based on the size of the array (say, from 1 to 20). But this would introduce an arbitrary limit on the size of expressions used :disappointed:

PR #107, opened 2019-03-13
fix scaling of coefficients in quadratic constraints

### LINKED ISSUE #106: Issue with (nonconvex) quadratic constraints
Consider the following:

```julia
using JuMP
using SCIP
using LinearAlgebra
m = Model(with_optimizer(SCIP.Optimizer))
@variable m x[1 : 3]
@variable m y[1 : 3]
@variable m z[1 : 3]
w = x × y - x × z
@constraint m w .== 1
optimize!(m)
```

SCIP reports that the problem is solved, but `value.(w)` results in

```julia
3-element Array{Float64,1}:
 1.9999999962747097
 1.9999999925494194
 1.9999999962747097
```

i.e., twice what it should be.

leethargo:
fix #106


leethargo:
This also passes (locally) the new tests in https://github.com/JuliaOpt/MathOptInterface.jl/pull/697.

PR #108, opened 2019-03-14
Handle parameter types properly.

### LINKED ISSUE #15: float64 parameters

tkoolen:
This aims to restore the same functionality that was introduced in https://github.com/SCIP-Interfaces/SCIP.jl/pull/16, but I chose to have the type of parameter be fully guided by what SCIP thinks it should be (using `SCIPparamGetType`), rather than using dispatch on the Julia side. `cconvert` should take care of the rest.

Fixes #15.

### DIFF
--- a/src/managed_scip.jl
+++ b/src/managed_scip.jl
@@ -49,9 +49,28 @@ function free_scip(mscip::ManagedSCIP)
     @assert mscip.scip[] == C_NULL
 end
 
-"Set generic parameter."
-function set_parameter(mscip::ManagedSCIP, name::String, value)
-    @SC SCIPsetParam(mscip, name, Ptr{Cvoid}(value))
+"Set a parameter."
+function set_parameter(mscip::ManagedSCIP, name::AbstractString, value)
+    param = SCIPgetParam(mscip, name)
+    if param == C_NULL
+        error("Unrecognized parameter: $name")
+    end
+    paramtype = SCIPparamGetType(param)
+    if paramtype === SCIP_PARAMTYPE_BOOL
+        @SC SCIPsetBoolParam(mscip, name, value)
+    elseif paramtype === SCIP_PARAMTYPE_INT
+        @SC SCIPsetIntParam(mscip, name, value)
+    elseif paramtype === SCIP_PARAMTYPE_LONGINT
+        @SC SCIPsetLongintParam(mscip, name, value)
+    elseif paramtype === SCIP_PARAMTYPE_REAL
+        @SC SCIPsetRealParam(mscip, name, value)
+    elseif paramtype === SCIP_PARAMTYPE_CHAR
+        @SC SCIPsetCharParam(mscip, name, value)
+    elseif paramtype === SCIP_PARAMTYPE_STRING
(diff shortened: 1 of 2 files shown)


leethargo:
Nice, thanks!

I wonder whether another level of safety could be introduced, where we compare the `typeof(value)` with the result of `SCIPparamGetType`, but I guess when users look up parameters in the SCIP documentation, they see the required type there anyways.

tkoolen:
I think `cconvert` adequately takes care of safety. This is what happens if you pass in a value of the wrong type:

```julia
SCIP.Optimizer(limits_gap="bla")

MethodError: Cannot `convert` an object of type String to an object of type Float64
Closest candidates are:
  convert(::Type{T<:Number}, !Matched::T<:Number) where T<:Number at number.jl:6
  convert(::Type{T<:Number}, !Matched::Number) where T<:Number at number.jl:7
  convert(::Type{T<:Number}, !Matched::Base.TwicePrecision) where T<:Number at twiceprecision.jl:250
  ...

Stacktrace:
 [1] cconvert(::Type, ::String) at ./essentials.jl:355
 [2] SCIPsetRealParam(::SCIP.ManagedSCIP, ::String, ::String) at /home/twan/.julia/packages/SCIP/DSWBY/src/wrapper/scip_param.jl:114
 [3] set_parameter(::SCIP.ManagedSCIP, ::String, ::String) at /home/twan/.julia/packages/SCIP/DSWBY/src/managed_scip.jl:66
 [4] set at /home/twan/.julia/packages/SCIP/DSWBY/src/MOI_wrapper.jl:111 [inlined]
 [5] (::getfield(SCIP, Symbol("##Optimizer#15#16")))(::Base.Iterators.Pairs{Symbol,String,Tuple{Symbol},NamedTuple{(:limits_gap,),Tuple{String}}}, ::Type) at /home/twan/.julia/packages/SCIP/DSWBY/src/MOI_wrapper.jl:42
 [6] (::getfield(Core, Symbol("#kw#Type")))(::NamedTuple{(:limits_gap,),Tuple{String}}, ::Type{SCIP.Optimizer}) at ./none:0
 [7] top-level scope at In[12]:1
```
which I think is clear enough.

leethargo:
Nice. Would you like a new tagged version? I think you should have the appropriate GitHub access rights to do that already.

tkoolen:
OK, done.

Issue #94, opened 2019-01-20
Expose callbacks for various SCIP plugin types.


leethargo:
Since SCIP support for plugins (callbacks) is more general that what is available in other MIP solvers, it makes sense to offer a thin layer around the SCIP style of them, for example:
- constraint handlers (lazy constraints)
- heuristics (based on LP solutions, but also from scratch type heuristics, or improving heuristics)
- relaxations (instead of, or in addition to LP)
- branching rules
- node selection rules

If/when a callback interface is established for MOI, this layer could be used then.

leethargo:
See also discussion in https://github.com/JuliaOpt/CPLEX.jl/pull/207 for a candidate interface that could be used as a template.

leethargo:
My personal motivation would be to expose constraint handler plugins first (in full generality) with a thin convenience wrapper (à la lazy constraint callback), in the coming weeks.

Then heuristics, and (much?) later the other plugin types.

mlubin:
I'd like to brainstorm how to expose constraint handler plugins in full generality without SCIP needing to depend on JuMP. Could you give a brief summary of what callbacks are needed to implement a constraint handler? I remember there are multiple functions to implement.

leethargo:
For constraint handlers, there are 3-4 [required callbacks](https://scip.zib.de/doc-6.0.1/html/CONS.php#CONS_FUNDAMENTALCALLBACKS):

- `check`: given a solution candidate, check constraint violation and return boolean 
- `enfo`(rce): given a solution candidate, iff constraint is violated, do something about it (e.g. add a constraint, create branches).
- `lock`: for each variable, say whether rounding up or down could violate this constraint. This could be set to "all true", conservatively, breaking some heuristics/preprocessing.

There are two versions of `enfo`, one for LP solutions, and one for pseudo-solutions (which have some different, heuristic origin and might not satisfy the LP relaxation), but it is common (I think) to use one implementation for both of these.

In addition, there are lots of [optional callbacks](https://scip.zib.de/doc-6.0.1/html/CONS.php#CONS_ADDITIONALCALLBACKS): Some of these are for data bookkeeping (memory mgmt, copies of data, transformation during presolve etc), others algorithmic (`sepa` for cut separation, `prop` for bounds propagation).

Looking at the documentation, I'm now hesitant with my previous claim of "full generality" ;-)
[...]

PR #109, opened 2019-03-24
add support for constraint handlers

rschwarz:
Adresses part of #94.


leethargo:
Very early sketches, still.

First idea is to have an `AbstractConstraintHandler` for which user define a subtype, which will also play the role of `SCIP_CONSHDLRDATA`.

I wanted to define methods (with nice Julia signatures) for the callbacks. But it's not quite clear yet how I will get a `@cfunction` pointer of these (after dispatch!). So, it might be better to use a `setcallback!(model, fun)` approach, like in MPB.

src/conshdlr.jl
```diff
@@ -0,0 +1,125 @@
...
+#   - DELAYPROP = -1
+#   - PROP_TIMING = SCIP_PROPTIMING_BEFORELP
+#   - PRESOLTIMING = SCIP_PRESOLTIMING_MEDIUM
+#   - MAXPREROUNDS = -1
+# - don't support optional methods: CONSHLDRCOPY, CONSFREE, CONSINIT, CONSEXIT,
+#   CONSINITPRE, CONSEXITPRE, CONSINITSOL, CONSEXITSOL, CONSDELETE, CONSTRANS,
+#   CONSINITLP, CONSSEPALP, CONSSEPASOL, CONSENFORELAX, CONSPROP, CONSPRESOL,
+#   CONSRESPROP, CONSACTIVE, CONSDEACTIVE, CONSENABLE, CONSDISABLE, CONSDELVARS,
+#   CONSPRINT, CONSCOPY, CONSPARSE, CONSGETVARS, CONSGETNVARS, CONSGETDIVEBDCHGS
+# - don't support linear or nonlinear constraint upgrading
```
mlubin: What is constraint upgrading?

src/conshdlr.jl
```diff
@@ -0,0 +1,125 @@
...
+
+# CONSENFOLP
+# (SCIP* scip, SCIP_CONSHDLR* conshdlr, SCIP_CONS** conss, int nconss,
+#  int nusefulconss, SCIP_Bool solinfeasible, SCIP_RESULT* result)
+
+function _consenfolp(scip::Ptr{SCIP_}, conshdlr::Ptr{SCIP_CONSHDLR},
+                     conss::Ptr{Ptr{SCIP_CONS}}, nconss::Cint,
+                     nusefulconss::Cint, solinfeasible::SCIP_Bool,
+                     result::Ptr{SCIP_RESULT})
+    return SCIP_OKAY
```
mlubin: Can you redirect to the user-provided callback by storing it in SCIP_CONSDATA?

src/conshdlr.jl
robertschwarz: Constraint upgrading means detecting special structure, and then converting a given constraint to a constraint of a different handler. For example, a user might create a linear constraint, but then SCIP detects that it is actually a set-cover constraint, so it will create a constraint of that type in order to use specialized methods for cuts, presolve etc. Similarly, a user might submit a constraint with a nonlinear, but then SCIP detects that it can be represented as a quadratic constraint handler and upgrades to that. Not sure whether the original constraints remain in the (transformed) problem or not.

src/conshdlr.jl
robertschwarz: Yes, that might work. Then we would have setter methods to store functions in the `AbstractConstraintHandler` subtype, I guess. Some of these are required, so it might be better to set them in the constructor already.

robertschwarz:
> You provide a top-level function that redirects to the user methods.

Yeah, we did something similar in the old SCIP.jl ([lazycb_wrapper](https://github.com/SCIP-Interfaces/SCIP.jl/blob/5338a7fb568bde80fb651820434a99238aa91cc9/src/mpb_interface.jl#L319-L339)).

In the IPOPT example, the user-defined Julia function is stored in the `prob` type, and then called as `prob.userfun(...)`. What I had imagined was that the user provides the methods that are then selected via dispatch, as in `userfun(prob, ...)`, using the specific subtype of `AbstractConstraintHandler`. But that approach might not actually have any benefits and might lead to errors with incomplete implementation of the callback interface.

robertschwarz:
I've also wondered whether to use `MOI.set` and `get` for access to the callbacks. But since this is SCIP specific, there is probably no benefit in that either.

rschwarz:
Have a bare-bones implementation, with a minimal (passing) test of a dummy constraint handler that does nothing (but the callbacks are being called).

Next step would be a more involved test/example that involves MOI, to see what kind of utility methods will be needed to make the mappings between `SCIP_CONS*` pointers and the related `SCIP_CONSDATA*` pointers, as well as conversions between C pointers and Julia objects.

I'm already worrying about using Julia objects (that can be GCed) as `SCIP_CONSHDLRDATA` and `SCIP_CONSDATA` and have the user be responsible to keeping them alive...

src/conshdlr.jl
```diff
@@ -0,0 +1,250 @@
...
+# enfoprio: integral has 0, nonlinear: -30 or less
+get(::AbstractConstraintHandler, ::EnforcePriority) = -15
+# checkprio: integral has 0, SCIP's plugins have -6000000 or more
+get(::AbstractConstraintHandler, ::CheckPriority) = -7000000
+
+# For semantics of callback methods, see SCIP documentation at:
+# https://scip.zib.de/doc-6.0.1/html/CONS.php#CONS_FUNDAMENTALCALLBACKS
+
+# possible return values:
+# SCIP_FEASIBLE: given solution is satisfies the constraint
```
mlubin: s/is satisfies/satisfies/ (below as well)

src/conshdlr.jl
```diff
@@ -0,0 +1,250 @@
...
+# SCIP_INFEASIBLE: given solution is violates the constraint
```
mlubin: s/is violates/violates/

mlubin:
> I'm already worrying about using Julia objects (that can be GCed) as SCIP_CONSHDLRDATA and SCIP_CONSDATA and have the user be responsible to keeping them alive...

Do you mean for the pure SCIP interface or for MOI? For MOI, could you create (and then free) these during the `optimize!` call?

rschwarz:
> Do you mean for the pure SCIP interface or for MOI? For MOI, could you create (and then free) these during the optimize! call?

I feel that it can affect both. I imagine that these objects are really created by the (MOI) user, just like you would create a `function, set` pair when you add a constraint. I guess that we should keep a copy/reference to it inside `SCIP.Optimizer`, to be sure.

A related issue is that the callback (on the SCIP/C side) receive a `SCIP*`. So, semantically, it would make sense that the `SCIP.jl` and `MOI` users would receive a `ManagedSCIP` and `SCIP.Optimizer`, respectively. I'm not sure whether these can be injected automatically, or we should tell users to store these references in their `ConstraintHandler` object.

By the way, I appreciate that you read the updated code, but I don't think it's ready for review, yet. I have not figured out what level of detail to expose to the SCIP.jl users and the variable naming is all over the place. For instance, I would like to make it clear whether a an argument is of type `Ptr{SCIP_CONS}`, `Ptr{SCIP_CONSDATA}`, or the corresponding user object (e.g. `DummyConstraint`).

I already started working on the example of a lazy linear constraint handler and it's not so obvious how to give the user access to the solution values. A `Dict{MOI.VI, Float64}`? Or a closure where they can fetch values for individual `MOI.VI`s? I guess I will not develop any further convenience on the pure SCIP or `ManagedSCIP` side and focus on `MOI` directly.

src/MOI_wrapper.jl
```diff
@@ -91,7 +91,7 @@ end
 
 "Go back from solved stage to problem modification stage, invalidating results."
 function allow_modification(o::Optimizer)
-    if SCIPgetStage(o) != SCIP_STAGE_PROBLEM
+    if !(SCIPgetStage(o) in [SCIP_STAGE_PROBLEM, SCIP_STAGE_SOLVING])
```
matbesancon:
```suggestion
    if !(SCIPgetStage(o) in (SCIP_STAGE_PROBLEM, SCIP_STAGE_SOLVING))
```

Will avoid allocating a vector

rschwarz:
So, I'm quite happy with this week's progress. While staying close to SCIP's API, one can now implement constraint handlers by defining 1-2 structs and implementing 2-4 methods (depending on the setting of `needs_constraints`).

There are two non-trivial examples in the tests now ([NaiveAllDiff](https://github.com/SCIP-Interfaces/SCIP.jl/pull/109/files#diff-0a54260d8d5bff0123e570a67ae086f9R85) by enumeration, [NoGoodCounter](https://github.com/SCIP-Interfaces/SCIP.jl/pull/109/files#diff-0a54260d8d5bff0123e570a67ae086f9R163) with "lazy" constraints) but I guess they will only make sense to people who understand constraint handlers at the SCIP level.

Still needs docstrings badly, but not today...

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+
+#
+# Fundamental callback functions with SCIP's C API.
+#
+# There is only a single, generic implementation for each of these, which are
+# passed to all user-defined SCIP constraint handlers, but they will call the
+# user's method in the function body.
+#
+
+"Generic `check` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `check` function, matching the signature from SCIP's C API.
"""
```

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+
+    # call user method via dispatch
+    res = check(constraint_handler, constraints, sol, checkintegrality,
+                checklprows, printreason, completely)
+    unsafe_store!(result, res)
+
+    return SCIP_OKAY
+end
+
+"Generic `enfolp` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `enfolp` function, matching the signature from SCIP's C API.
"""
```

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+
+    # call user method via dispatch
+    res = enforce_lp_sol(constraint_handler, constraints, nusefulconss,
+                         solinfeasible)
+    unsafe_store!(result, res)
+
+    return SCIP_OKAY
+end
+
+"Generic `enfops` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `enfops` function, matching the signature from SCIP's C API.
"""
```

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+
+    # call user method via dispatch
+    res = enforce_pseudo_sol(constraint_handler, constraints, nusefulconss,
+                             solinfeasible, objinfeasible)
+    unsafe_store!(result, res)
+
+    return SCIP_OKAY
+end
+
+"Generic `lock` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `lock` function, matching the signature from SCIP's C API.
"""
```

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+
+
+#
+# Additional callback functions for memory management.
+#
+# The implementation should work for all constraint handlers defined in Julia,
+# so there is no method for the user to implement.
+#
+
+"Generic `free` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `free` function, matching the signature from SCIP's C API.
"""
```

src/conshdlr.jl
```diff
@@ -0,0 +1,394 @@
...
+    # Here, we should free the constraint handler data. But because this is an
+    # object created and owned by Julia, we will let GC do it.
+    # Instead, we will just set the pointer to NULL, so that SCIP will think
+    # that it is taken care of.
+    SCIPconshdlrSetData(conshdlr, C_NULL)
+
+    return SCIP_OKAY
+end
+
+"Generic `delete` function, matching the signature from SCIP's C API."
```
matbesancon:
```suggestion
"""
Generic `delete` function, matching the signature from SCIP's C API.
"""
```

src/convenience.jl
```diff
@@ -0,0 +1,13 @@
+## Related to constraint handlers
+
+"Recover Julia object which is attached to user constraint (via constraint data)."
```
matbesancon:
```suggestion
"""
Recover Julia object which is attached to user constraint (via constraint data).
"""
```

src/convenience.jl
```diff
@@ -0,0 +1,13 @@
...
+function user_constraint(cons_::Ptr{SCIP_CONS})
+    @assert cons_ != C_NULL
+    consdata_::Ptr{SCIP.SCIP_CONSDATA} = SCIPconsGetData(cons_)
+    return unsafe_pointer_to_objref(consdata_)
+end
+
+"Extract solution values for given variables."
```
matbesancon:
```suggestion
"""
Extract solution values for given variables.
"""
```

src/managed_scip.jl
```diff
@@ -15,6 +15,8 @@ mutable struct ManagedSCIP
     conss::Dict{ConsRef, Ref{Ptr{SCIP_CONS}}}
     var_count::Int64
     cons_count::Int64
+    conshdlrs::Dict{Any, Ptr{SCIP_CONSHDLR}}
+    conshdlrconss::Dict{Any, Ptr{SCIP_CONS}}
```
matbesancon: Are these dict indexed by anything or by a string / int / something?

rschwarz: Well, they are indexed by the user-defined types, so, subtypes of `SCIP.AbstractConstraintHandler` or `SCIP.AbstractConstraint{CH}` respectively. But it can be mixed, so I don't think it's possible to use a type parameter.

rschwarz:
OK, I have now done everything that I had planned to do, so I consider it ready for review.

Before merging it, I would like to test it with a real-word usage case ([PipeLayout.jl](https://github.com/rschwarz/PipeLayout.jl)) first, which will take some time.

rschwarz:
I was pondering on changing the MOI-facing function [include_conshdlr](https://github.com/SCIP-Interfaces/SCIP.jl/pull/109/files#diff-11f96aa0021aedc181302287c5e65582R27) to use `MOI.submit`.
This would probably entail using another `struct` (a `Submittable`) that holds the actual `AbstractConstraintHandler` in addition to the keyword arguments of the current functions.

But it's not clear what the benefits are, for the user. Would it be easier to `submit` the plugin from JuMP? Also, it might be confusing, given that the discussion in https://github.com/JuliaOpt/MathOptInterface.jl/pull/782 points to using `SolverAttribute`s for callbacks instead, for most other MIP solvers out there.

rschwarz:
So, this works well enough for my use-case ([treetopohdlr.jl](https://github.com/rschwarz/PipeLayout.jl/blob/62a58d2793ef56b7a97895e134709ec2fd6a5bb8/src/models/gndstruct_discdiam/treetopohdlr.jl), [semisubhdlr.jl](https://github.com/rschwarz/PipeLayout.jl/blob/62a58d2793ef56b7a97895e134709ec2fd6a5bb8/src/models/gndstruct_discdiam/semisubhdlr.jl), [cbtopo.jl](https://github.com/rschwarz/PipeLayout.jl/blob/62a58d2793ef56b7a97895e134709ec2fd6a5bb8/src/models/gndstruct_discdiam/cbtopo.jl)) that I would like to merge & release at this point, if there are no objections or questions.

mlubin:
Seems reasonable to merge this and look at proper support from JuMP in a separate PR.

rschwarz:
>  look at proper support from JuMP in a separate PR

You mean in a solver-independent way?

I find the [way I create constraints from callbacks](https://github.com/rschwarz/PipeLayout.jl/blob/62a58d2793ef56b7a97895e134709ec2fd6a5bb8/src/models/gndstruct_discdiam/treetopohdlr.jl#L57) nice enough, based [thinly](https://github.com/rschwarz/PipeLayout.jl/blob/62a58d2793ef56b7a97895e134709ec2fd6a5bb8/src/util.jl#L68) on `JuMP.@build_constraint`.

mlubin:
@rschwarz, just in a documented way that's usable without PipeLayout.jl, not solver-independent :)

rschwarz:
> just in a documented way that's usable without PipeLayout.jl

Good point. Maybe this calls for a tested example, as another JuMP-dependent test stage just like MINLPTests.

matbesancon:
> Maybe this calls for a tested example, as another JuMP-dependent test stage just like MINLPTests.

Or even in an example folder called from the normal travis build (if small enough that should be fine with build time)

rschwarz:
> Or even in an example folder called from the normal travis build (if small enough that should be fine with build time)

But then we have the problem again that we can not upgrade SCIP to a new version of MOI as long as a compatible JuMP is not released.

PR #110, opened 2019-03-29
Add OSX support (+ Travis testing)

tkoolen:
Took about 5 minutes for the actual change and several hours to get Travis working. I initially wanted to use the precompiled binaries, but I ran into an illegal instruction on OSX, https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/512838942#L227, and decided to compile from source.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,6 @@ matrix:
   allow_failures:
     - julia: nightly
 addons:
-  homebrew:
-    packages:
-      - gcc@6
   apt:
     packages:
     - libblas3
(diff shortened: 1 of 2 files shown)


leethargo:
Great, thanks a lot!

> Took about 5 minutes for the actual change and several hours to get Travis working.

I know that feel :disappointed: 

> I initially wanted to use the precompiled binaries, but I ran into an illegal instruction on OSX.

Too bad. It looks like the issue could be related to a different version of GMP found at Travis.


I saw that you build SCIP without IPOPT (which would be nasty to do manually) and I was worried that the MINLP tests might fail, or take forever to solve, but that stage is still Linux-only, so it's all good :+1:

PR #113, opened 2019-04-03
MOI-level support for indicator constraints

matbesancon:


leethargo:
Awesome, thanks!

One comment: It seems that SCIP fails when the indicator variable is not already binary. Could we add a check for that, similar to the [bounds check in SOC constraints](https://github.com/SCIP-Interfaces/SCIP.jl/blob/master/src/MOI_wrapper/soc_constraints.jl#L8-L13)? Then, SCIP.jl would be spared (parts of) the nasty stacktrace.

src/MOI_wrapper/indicator_constraints.jl
```diff
@@ -0,0 +1,35 @@
...
+    a::Vector{T}
+    rhs::T
+end
+
+MOI.supports_constraint(::Optimizer, ::Type{VECTOR}, ::Type{IS}) where {IS <: IndicatorSet{<:AbstractFloat}} = true
+
+function MOI.add_constraint(o::Optimizer, func::VECTOR, set::IndicatorSet{Float64})
+    length(func.variables) == length(set.a) + 1 || throw(DimensionMismatch("Indicator set and VectorOfVariables length mismatch"))
+    allow_modification(o)
+    y = VarRef(func.variables[1].value)
```
leethargo: This would be a nice place to check whether y is already binary.

test/MOI_additional.jl
```diff
@@ -107,6 +107,28 @@ end
     x = MOI.add_variable(optimizer)
     t = MOI.add_constraint(optimizer, x, MOI.ZeroOne())
     @test_throws ErrorException b = MOI.add_constraint(optimizer, x, MOI.Interval(2.0, 3.0))
+
+    MOI.empty!(optimizer)
```
leethargo: Nitpick: These tests make sense, but are not really about binary variables or implicit bounds.
Could you move them down with the other indicator tests?

matbesancon: sure yup

src/MOI_wrapper/indicator_constraints.jl
matbesancon: shouldn't we do it in the SCIP part instead of MOI? (meaning here in the parts defined in `managed_scip.jl`?)

leethargo: Either way is fine with me. I only looked up how I did it with SOC but can not remember the rationale. I guess I thought that the MOI part was more "user-facing", but of course if we move the check "down" the call stack, it will cover more cases!

PR #115, opened 2019-04-03
Check SCIP stage before result queries.

### LINKED ISSUE #114: segfault when not optimized
One gets a nasty segfault without any helpful information (from SCIP nor Julia) when getting variables without having optimized. 

MWE: https://github.com/SCIP-Interfaces/SCIP.jl/blob/090911ee1c5a2306272025129f1da5fc13f0c10d/test/MOI_additional.jl#L403

If line `MOI.optimize!` is removed

leethargo:
Fixes #114

### DIFF
--- a/test/MOI_additional.jl
+++ b/test/MOI_additional.jl
@@ -439,9 +439,9 @@ end
     @test_throws ErrorException MOI.get(optimizer, MOI.ConstraintPrimal(), c)
     @test_throws ErrorException MOI.get(optimizer, MOI.ObjectiveBound())
     @test_throws ErrorException MOI.get(optimizer, MOI.RelativeGap())
-    @test_throws ErrorException MOI.get(optimizer, MOI.SolveTime())
+    @test MOI.get(optimizer, MOI.SolveTime()) ≈ 0.0 atol=atol rtol=rtol
     @test_throws ErrorException MOI.get(optimizer, MOI.SimplexIterations())
-    @test_throws ErrorException MOI.get(optimizer, MOI.NodeCount())
+    @test MOI.get(optimizer, MOI.NodeCount()) == 0
 
     # after optimize
     MOI.optimize!(optimizer)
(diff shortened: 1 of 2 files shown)


test/MOI_additional.jl
```diff
@@ -412,3 +412,49 @@ end
...
+    @test MOI.get(optimizer, MOI.PrimalStatus()) == MOI.NO_SOLUTION
+
+    @test_throws ErrorException MOI.get(optimizer, MOI.ObjectiveValue())
+    @test_throws ErrorException MOI.get(optimizer, MOI.VariablePrimal(), x)
+    @test_throws ErrorException MOI.get(optimizer, MOI.ConstraintPrimal(), c)
+    @test_throws ErrorException MOI.get(optimizer, MOI.ObjectiveBound())
+    @test_throws ErrorException MOI.get(optimizer, MOI.RelativeGap())
+    @test_throws ErrorException MOI.get(optimizer, MOI.SolveTime())
+    @test_throws ErrorException MOI.get(optimizer, MOI.SimplexIterations())
+    @test_throws ErrorException MOI.get(optimizer, MOI.NodeCount())
```
matbesancon: Node count can only be queried at the end of optimization? (not usable in callbacks for instance?)

leethargo: Yeah, actually it's not necessary here, same for `SolveTime`. But it is necessary for `SimplexIterations`, it seems, so I just thought that maybe SCIP was too lax in its assertions.

But now that you mention it, it does make sense to call some of these during the solve process.
Will re-check docs.

matbesancon:
In `MathProgBase` if I recall correctly, variables were yielding `NaN` in that case (with a warning the first time you access them)

leethargo:
OK, that is also a viable option. But the [MOI docs](http://www.juliaopt.org/MathOptInterface.jl/stable/apimanual/#Solving-and-retrieving-the-results-1) actually say:
> If PrimalStatus is not NO_SOLUTION, then the primal may be retrieved with the VariablePrimal attribute:

So in a sense, the users are breaking the MOI contract by asking for the solution values unconditionally.

matbesancon:
Ah yes fair point, an error might be the most helpful then :+1:

PR #117, opened 2019-04-03
add dimension method for Indicator

matbesancon:
guess who forgot to add a method definition for `IndicatorSet`


leethargo:
Is this needed for JuMP?
The MOI tests worked fine without it.

matbesancon:
yes it seems not to be tested / enforced but is in the MOI interface for sets

PR #118, opened 2019-04-09
activate more QCP MOI tests

rschwarz:
(requires new release of MOI)


matbesancon:
shouldn't MOI add a new minor version for added features? It's a bit awkward to require a specific patch (and I think adds upper bounds)

leethargo:
> shouldn't MOI add a new minor version for added features?
It's not really a new feature, just an extended test suite. Of course, I was just guessing the next MOI version after I realized that these tests are just on master.

> It's a bit awkward to require a specific patch (and I think adds upper bounds)
This is something I don't know about. I thought that for upper bounds, one would have to specify a second version.

In any case, this would now be specified in the Project.toml, which have different syntax again.

PR #119, opened 2019-04-09
Add Project.toml, rm REQUIRE

rschwarz:


leethargo:
Should merge #118 before, PR created prematurely :confused:

rschwarz:
Maybe this should be rebased on master, because it currently (accidentally) contains the commits from #118 as well. #118 should only be merged after a new MOI release (v0.9?) was tagged, which I think didn't happen yet.

matbesancon:
this would be practical indeed

rschwarz:
@matbesancon your `Project.toml` from #120 looks different from the one I created. I don't actually remember what process I followed, but I see that I'm missing the test dependencies, for example. Feel free to apply those changes as you see fit.

Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "SCIP"
+uuid = "82193955-e24f-5292-bf16-6f2c5261a85f"
+
```
matbesancon: I think the version is always required and must corresponds to the version yet to be released (not released yet)

Project.toml
```diff
@@ -0,0 +1,13 @@
...
+[deps]
+CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[compat]
+CEnum = "≥ 0.1.0"
+MathOptInterface = "≥ 0.8.4"
+julia = "≥ 1.0.0"
```
matbesancon: we should add:

```

[extras]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"

[targets]
test = ["Test"] 
```

Project.toml
```diff
@@ -0,0 +1,13 @@
+name = "SCIP"
+uuid = "82193955-e24f-5292-bf16-6f2c5261a85f"
+
+[deps]
+CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
matbesancon:
```suggestion
```

matbesancon: Test is not needed by the package itself, just in the tests

Project.toml
rschwarz: I see. And dropping `REQUIRE` would be considered a breaking change, I guess.

matbesancon: yes I'd say so. Also, beware of `>=` compat, it means no upper-bound, which is problematic for breaking changes expected (MOI, CEnum next releases could break everything)

rschwarz:
Thanks for your comments. I believe the first TOML file was generated from the REQUIRE by a script that was shared on Discourse. But that doesn't mean that the version spec will be any good, if there were no upper bounds in the REQUIRE.

Not sure if this should be merged now, to register a release, or better wait on MOI's update to v0.9.

matbesancon:
we can register a last release patch, given the fix for Windows, it would be interesting

PR #121, opened 2019-05-07
Indicator moi

matbesancon:
changing indicator to MOI version. Requires Project.toml merged first
Only works for MOI master, and soon 0.9 when released


tkoolen:
Thank you for working on this stuff. I see that SCIP also supports indicator constraints over arbitrary constraint types ([super indicator constraints](https://scip.zib.de/doc/html/cons__superindicator_8h.php)) and that `MOI.IndicatorSet` is also designed to potentially support that. Just wondering: is supporting those kinds of constraints also on your radar?

rschwarz:
I think that supporting `cons_superindicator` should be possible, but would require either some code duplication or refactoring of the code in the `ManagedSCIP` part and `MOI` wrapper: Right now, the existing functions related to constraints all do three things: create constraint (data), add constraint to problem, and register constraint for reference counting. To add a constraint as part of super indicator, we would only need the first part, I guess.

matbesancon:
I think it's not the same SCIP constraint if you have arbitrary constraints, and MOI is meant for linear constraints I believe, including `>=, <=, ==`

PR #122, opened 2019-05-08
Add Windows support

iSoron:
The package currently fails to build on Windows with the message `SCIP is currently not supported on "NT"`. This pull request fixes this by modifying `deps/build.jl` to search for the correct DLL. I have verified not only the package builds successfully with these changes, but also that all unit tests pass. 

The function `write_depsfile` needed to be modified because, on Windows, the `path` variable contains backslashes which need to be quoted.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,30 @@
+environment:
+  matrix:
+  - julia_version: 1
+  SCIPOPTDIR: C:\scipoptdir
+
+platform:
+  - x64 # 64-bit
+
+branches:
+  only:
+    - master
+
+notifications:
+  - provider: Email
+    on_build_success: false
+    on_build_failure: false
+    on_build_status_changed: false
+
+install:
+  - ps: iex ((new-object net.webclient).DownloadString("https://raw.githubusercontent.com/JuliaCI/Appveyor.jl/version-1/bin/install.ps1"))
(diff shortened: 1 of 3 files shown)


rschwarz:
Cool, thanks!

It would be great if we could also set up appveyor to run automated tests on Windows.
I have not done that before, but could look into that later.

matbesancon:
Maybe we can add it during that PR right away (never done it either)

matbesancon:
The PR needs to add an appveyor file as in:
https://github.com/JuliaLang/Example.jl/blob/master/.appveyor.yml

matbesancon:
@rschwarz I think there are some custom steps because of the way SCIP is imported, you know them better than me

rschwarz:
> Maybe we can add it during that PR right away.

Yes, that's what I meant.

> I think there are some custom steps because of the way SCIP is imported

Yes, SCIP needs to be downloaded and installed. We also need to set the `SCIPOPTDIR` variable.

rschwarz:
@fserra, @mattmilten I've tried to setup AppVeyor integration for SCIP.jl, by installing it to the SCIP-Interfaces group, but only to "selected projects", namely SCIP.jl. If I have disabled the integration for PySCIPOpt through this action (it does not show up in the settings), please let me know or change it back.

mattmilten:
I guess back when I added appveyor, I only added it for PySCIPOpt.

rschwarz:
> I guess back when I added appveyor, I only added it for PySCIPOpt.

That's alright. My concern is that I might *disabled* it for PySCIPOpt inadvertantly.

PR #123, opened 2019-05-10
Add user-friendly interface for statistics printing functionality.

tkoolen:
Allows you to do e.g.

```julia
print_timing_statistics(mscip)
```

or

```julia
print_timing_statistics(mscip, filename)
```

The docstrings for these functions look like

```julia
help?> SCIP.print_timing_statistics
  SCIP.print_timing_statistics(mscip::ManagedSCIP)
  SCIP.print_timing_statistics(mscip::ManagedSCIP, file_descriptor::RawFD)
  SCIP.print_timing_statistics(mscip::ManagedSCIP, stream::Union{Base.Filesystem.File, IOStream})
  SCIP.print_timing_statistics(mscip::ManagedSCIP, filename::AbstractString)

  Print statistics (calls SCIP.SCIPprintTimingStatistics).

  The one-argument version prints to standard output.

  Note that the method that takes a stream only applies to synchronous Files and IOStreams. In particular, it doesn't apply to stdout.
```

### DIFF
--- a/src/managed_scip.jl
+++ b/src/managed_scip.jl
@@ -276,3 +276,56 @@ function add_indicator_constraint(mscip::ManagedSCIP, y, x, a, rhs)
     @SC SCIPaddCons(mscip, cons__[])
     return store_cons!(mscip, cons__)
 end
+
+# Transform SCIP C function name as follows:
+# 1. Remove leading "SCIP" part (drop the first four characters).
+# 2. Convert camel case to snake case.
+# For example, `SCIPprintStatusStatistics` becomes `print_status_statistics`.
+const STATISTICS_FUNCS = map(x -> Symbol(camel_case_to_snake_case(string(x)[5 : end])), SCIP_STATISTICS_FUNCS)
+
+for (scip_statistics_func, statistics_func) in zip(SCIP_STATISTICS_FUNCS, STATISTICS_FUNCS)
+    @eval begin
+        """
+            $($statistics_func)(mscip::ManagedSCIP)
+            $($statistics_func)(mscip::ManagedSCIP, file_descriptor::RawFD)
+            $($statistics_func)(mscip::ManagedSCIP, stream::Union{Base.Filesystem.File, IOStream})
+            $($statistics_func)(mscip::ManagedSCIP, filename::AbstractString)
+
+        Print statistics (calls `$($scip_statistics_func)`).
+
+        The one-argument version prints to standard output.
+
+        Note that the method that takes a stream only applies to synchronous
+        `File`s and `IOStream`s. In particular, it doesn't apply to `stdout`.
+        """
+        function $statistics_func end
+
+        function $statistics_func(mscip::ManagedSCIP)
+            ret = $scip_statistics_func(mscip, C_NULL)
+            ret !== nothing && @assert ret == SCIP_OKAY
+            return nothing
+        end
+
+        function $statistics_func(mscip::ManagedSCIP, file_descriptor::RawFD)
+            file = Libc.FILE(file_descriptor, "w")
+            try
+                ret = $scip_statistics_func(mscip, file)
+                ret !== nothing && @assert ret == SCIP_OKAY
+            finally
+                close(file)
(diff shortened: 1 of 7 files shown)


src/managed_scip.jl
```diff
@@ -276,3 +276,50 @@ function add_indicator_constraint(mscip::ManagedSCIP, y, x, a, rhs)
...
     return store_cons!(mscip, cons__)
 end
+
+# Transform SCIP C function name as follows:
+# 1. Remove leading "SCIP" part (drop the first four characters).
+# 2. Convert camel case to snake case.
+# For example, `SCIPprintStatusStatistics` becomes `print_status_statistics`.
+const STATISTICS_FUNCS = map(x -> Symbol(camel_case_to_snake_case(string(x)[5 : end])), SCIP_STATISTICS_FUNCS)
+
+for (scip_statistics_func, statistics_func) in zip(SCIP_STATISTICS_FUNCS, STATISTICS_FUNCS)
```
matbesancon: shouldn't this part also be a macro?

tkoolen: This is how it's done in Base, e.g. https://github.com/JuliaLang/julia/blob/f0ffb29feb0499ce49d621acfa8a2ce192e8dba0/base/operators.jl#L524-L534.

matbesancon: fair point

tkoolen:
Not sure what's going on on Windows. Any ideas?

tkoolen:
Did a bunch of debugging.

* writing to a raw file descriptor on Windows with `fputs` works just fine, i.e.
```julia
filename = "statistics.txt"
open(filename, "w") do io
    rawfd = RawFD(fd(io))
    file = Libc.FILE(rawfd, "w")
    ccall(:fputs, Cvoid, (Ptr{Cchar}, Ptr{Libc.FILE}), "bla", file)
    close(file)
end
```
* the `ReadOnlyMemoryError`s still appear if I just use `RawFD(1)` (i.e., `stdout`) as the file descriptor (this also just works on Linux/OSX)
* the one-argument versions of the statistics printing functions (which pass `C_NULL` as the second argument to the SCIP functions and print to `stdout`) *do* work.

I'm a little stuck at this point.

rschwarz:
Sorry, I have not worked with Windows or `RawFD`, so I would not know about any gotchas. Here are some ideas that come to mind (but seem unlikely causes to me):

- I manually defined `const FILE = Cvoid` in `manual_commons.jl`.
- This (or something else) might lead to an incorrent verison of `unsafe_convert` to be applied in the Clang-generated `ccall` wrappers.
- SCIP itself (if you follow the calls from `SCIPprintStatusStatus` to `logMessage` in `message_default.c`) call both `fputs` and `fflush`.

Search for "ReadOnlyMemoryError ccall windows" does yield some issues on GitHub, so it might also be a problem upstream.

A "work-around" would be to only implement and test the functions on non-Windows platforms :disappointed:

tkoolen:
Thanks very much for the suggestions.

I tried using `FILE` from `Base.Libc` instead of defining it as `Cvoid`; unfortunately it didn't fix the problem.

> A "work-around" would be to only implement and test the functions on non-Windows platforms 😞

Yeah, and at least the method that prints to standard output (by passing `C_NULL` as the second argument) works on Windows.

tkoolen:
Alright, at least for now I've removed the methods that rely on passing `Ptr{FILE}`, leaving only the one that passes `C_NULL` and prints to `stdout`. This works on Windows, and if people want to save the output to a file instead, they can always just use `redirect_stdout`, as I did in the test.

rschwarz:
> if people want to save the output to a file instead, they can always just use redirect_stdout, as I did in the test.

Sounds good to me. Do you want to merge it as-is? We could register a new version afterwards.

tkoolen:
Yep, ready to go from my perspective. I'll merge. If you don't mind, could you tag a new version? I'm not used to the new package registration / version tagging setup yet, and I'd like to experiment with my own repositories first.

PR #125, opened 2019-05-16
Update MINLPTests to v0.3

rschwarz:
This is now registered.


rschwarz:
See https://github.com/JuliaOpt/MINLPTests.jl/pull/8.

PR #126, opened 2019-05-17
Prepare for MOI v0.9

rschwarz:
see https://github.com/JuliaOpt/MathOptInterface.jl/issues/736


rschwarz:
This should be ready (passes the tests locally).

In fact, most of the changes in this commit are about implementing features that existed in MOI prior to v0.9, but that I learned about via the newly added tests and docs :man_student:

rschwarz:
If we want to pass the tests in CI here, we should also change the version requirements in the `project.toml`, I think. I did not do so before, because the MOI version was not released yet.

matbesancon:
From the failure, it looks like IpOpt needs to be upgraded first?

rschwarz:
I don't know where you see something specific to IPOPT. The first failing log that I looked at in Travis only shows MOI unit test stuff.

In any case, SCIP.jl does not use IPOPT.jl, so it should not matter?

matbesancon:
> I don't know where you see something specific to IPOPT.

Log on Travis:
https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/571451684#L318

Anyway the error seems linked to something else

rschwarz:
> Log on Travis:

Yeah, but message from Ipopt always appears (once), even when everything passes. So it does not mean anything.

rschwarz:
OK, the failing tests seem to point to real problems in SCIP, not an out-of-date `Project.toml`. Probably related to changes (tests) that were added to MOI after this PR was created. I will need to take a closer look later.

rschwarz:
Unfortunately, the MINLPtests are failing with a `Pkg` dependency problem that I can't quite figure out from the [error message](https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/571798343#L329).

matbesancon:
It seems to require JuMP, which has not been upgraded yet to MOI 0.9

matbesancon:
@rschwarz there was a problem with the way the compats were specified, on #121 CEnum v0.2 is not compatible with the way we use it

rschwarz:
So, we could merge this to `master` despite the failing `MINLPTests` (waiting for JuMP release, waiting for Julia bugfix), so that we can take care of the other open PRs.

But I don't know whether we should tag a new version for SCIP.jl before a compatible JuMP exists?

matbesancon:
> So, we could merge this to master despite the failing MINLPTests

Maybe we can add an allowed failure on the travis script?

> But I don't know whether we should tag a new version for SCIP.jl before a compatible JuMP exists?

I think we should, officially, JuMP is just one of the interfaces on top of MOI, this should not be blocking for solvers

rschwarz:
Good points. I'm not so clear about the Travis syntax on allowing that stage to fail. First attempt incoming...

PR #130, opened 2019-08-28
simplify usage of MINLPTests (override status codes globally)

rschwarz:
Requires new release of MINLPTests with https://github.com/JuliaOpt/MINLPTests.jl/pull/20 merged.


rschwarz:
In addition to the changes so far, we could also switch from the explicit `include` to an `exclude` with rationales :man_shrugging:.

PR #135, opened 2019-10-18
Add access to dual solution (for LP problems)

rschwarz:
This requires that a lot of SCIP settings be disabled, so we should also provide some utilities around that (setting emphasis etc.).

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -194,6 +194,35 @@ function MOI.optimize!(o::Optimizer)
     return nothing
 end
 
+# Some utilities:
+
+"Make sure that SCIP is currently in one of the allowed stages."
+function assert_stage(o::Optimizer, stages)
+    stage = SCIPgetStage(o)
+    if !(stage in stages)
+        error("SCIP is wrong stage ($stage, need $stages), can not query results!")
+    end
+end
+
+"Make sure that the problem was solved (SCIP is in SOLVED stage)."
+function assert_solved(o::Optimizer)
+    # SCIP's stage is SOLVING when stopped by user limit!
+    assert_stage(o, (SCIP_STAGE_SOLVING, SCIP_STAGE_SOLVED))
+
+    # Check for invalid status (when stage is SOLVING).
+    status = SCIPgetStatus(o)
+    if status in (SCIP_STATUS_UNKNOWN,
+                  SCIP_STATUS_USERINTERRUPT,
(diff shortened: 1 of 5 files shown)


fserra:
I wouldn't put effort into this. Accessing the dual solutions in SCIP is super buggy. Even if you disable everything. If I remember correctly if you add bounds as constraints SCIP will anyway remove them. Then SCIP tries to rebuild the dual multipliers from the lp solver, but it doesn't realize if there are repeated constraints and so it will assign the same multiplier to different constraints, which is just wrong.

rschwarz:
> I wouldn't put effort into this. Accessing the dual solutions in SCIP is super buggy. 
Well, thanks for the warning :sob: 

I looked at [PySCIPOpt/#136](https://github.com/SCIP-Interfaces/PySCIPOpt/issues/136) and the related changes and thought that the problems are fixed now. In the Python code, there is a check for SCIP's version > 6.0, then use the `SCIPgetDualSolVal` and inside that function, there is special code for the "bound constraints".

So, most of the MOI tests about dual values actually pass. But those that fail, do so for different reasons. I think for one of them, the status changes: That is, with default settings, SCIP detects correctly that the problem is unbounded. But with presolve etc disabled, it says that it's optimal or something.

I have some interest in this beyond SCIP.jl, because I'd hoped to use this also in LITIC, but I guess we have to continue building an auxiliary LP just for the dual solution...

fserra:
yeah Matthias tried to fix it somehow, but I think that if you add constraints like `x <= 1`, `x<=1`, `x<=1`  then you should get something wrong. I can imagine that no test is that creative ;)

rschwarz:
So, the proper solution is to extract the LP, and ask the LP solver directly?
Or better, to build the problem with the LP solver, so that SCIP does no presolve...

fserra:
build it with the lp so that scip has nothing to do with it ;)

the-oaks:
@fserra may I ask if this is an issue with scip or their default liner solver soplex? (I apologize if I am asking a question that implied from the conversation already), thank you very much!

rschwarz:
The issue is with SCIP itself. If I understand it correctly, it can't even be fixed easily, because of the design based on generic constraints.

fserra:
I think that the problem is that SCIP does some presolving even if presolve is off

the-oaks:
@rschwarz @fserra thanks a lot for the explanation! cheers

PR #136, opened 2019-10-23
Project.toml for gen folder

matbesancon:
I'm not sure CEnum is really needed for the gen part


rschwarz:
I don't think it's needed to run the generation script. That `Project.toml` was simple copy-pasted from the environment I use for SCIP.jl dev.

PR #137, opened 2019-10-25
Fix indicator

matbesancon:
Adding the constant term to indicator constraints


rschwarz:
Do you want to merge it, so you get the fix, or wait for new MOI tag? Either way is fine with me.

matbesancon:
we can merge it now since it is self-contained (and a bug fix kind of needed), once the new MOI is released we can remove the test function

rschwarz:
OK, go ahead, then.

Issue #142, opened 2019-11-08
Building SCIP does not locate dylib although SCIPOPTDIR is set


jeremyomer:
Hello, I am using MacOs 10.12 and I just installed SCIP from scip.zib.de using  SCIPOptSuite-6.0.1-Darwin.sh file. I also properly set the SCIPOPTDIR environment to point to the install directory. I encountered the same very weird issue when trying to build SCIP.jl with Julia 1.1 and 1.2. In both cases I get the same error:

```
(v1.1) pkg> build SCIP
  Building SCIP → ~/.julia/packages/SCIP/IJWEZ/deps/build.log
┌ Error: Error building `SCIP`:
│ ERROR: LoadError: Unable to locate SCIP installation.
│ Tried:
│ 	/Users/jeremy/Applications/SCIPOptSuite-6.0.1-Darwin/bin/libscip.dylib
│ 	/Users/jeremy/Applications/SCIPOptSuite-6.0.1-Darwin/lib/libscip.dylib
│ 	libscip.dylib
│ Note that this must be downloaded separately from scip.zib.de.
```

But, the required library is present at one of the tried location:
```
[jeremy ~/Applications ] ls /Users/jeremy/Applications/SCIPOptSuite-6.0.1-Darwin/lib/
cmake                 libgcg.3.0.dylib      libscip.6.0.1.0.dylib libscip.dylib         libsoplex.a           libzimpl.a
libgcg.3.0.1.0.dylib  libgcg.dylib          libscip.6.0.dylib     libsoplex-pic.a       libzimpl-pic.a
```
I'd really appreciate help to start using SCIP in my research. Thanks

rschwarz:
I don't have a machine with OS X available, so I can't really help with that, sorry.

The error message does look familiar though. I think we had the same issue when we tried to use precompiled binaries for Travis CI. That's why SCIP is still [compiled](https://github.com/SCIP-Interfaces/SCIP.jl/blob/master/.travis.yml#L32) on Travis for OS X.

Maybe it's a file permissions problem, or something like that?

rschwarz:
I get the same error when using the precompiled OS X in Travis (see #143).

Maybe you can try compiling SCIP from source? This works in Travis.

jeremyomer:
Thank you for the answer ! 

Indeed, it worked by compiling SCIP from source with CMAKE. I'm closing the issue since another one was open on the subject.

PR #143, opened 2019-11-11
travis: try to use precompiled SCIP on OS X

rschwarz:


rschwarz:
[Travis can not find the `libscip.dylib`](https://travis-ci.org/SCIP-Interfaces/SCIP.jl/jobs/610513471#L187) of the precompiled SCIP:
```
┌ Error: Error building `SCIP`: 
│ ERROR: LoadError: Unable to locate SCIP installation.
│ Tried:
│ 	/Users/travis/build/SCIP-Interfaces/SCIP.jl/install/bin/libscip.dylib
│ 	/Users/travis/build/SCIP-Interfaces/SCIP.jl/install/lib/libscip.dylib
│ 	libscip.dylib
```
Just like in #142.

fserra:
interesting. As a patch, maybe you can try adding the library path to `DYLD_LIBRARY_PATH`

rschwarz:
> As a patch, maybe you can try adding the library path to DYLD_LIBRARY_PATH

This was not necessary before, when building SCIP from source. But feel free to try stuff :-) You should be able to push to this branch.

The current travis script does not print the actual contents of `install/lib`, but `libscip.dylib` should be there (I extracted locally).

rschwarz:
In PR #110, when using the precompiled binaries on OS X, there was an illegal instruction in SoPlex/GMP. Maybe this is fixed with 6.0.2, or maybe some related problem is causing `dlopen` to fail.

PR #145, opened 2019-12-19
Add example of SCIP with JuMP

mrchaos:
all test passed

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,7 +8,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 
 [compat]
-CEnum = "^0.1.0"
+CEnum = "^0.1.0, 0.2"
 MathOptInterface = "^0.9.6"
 julia = "^1.0.0"
 
--- a/example/scip_with_jump_minlp.jl
+++ b/example/scip_with_jump_minlp.jl
@@ -1,0 +1,46 @@
+using JuMP
+#using JuMP:MOI
+using SCIP
+
+function solve_model(m::Model;mixed::Bool = false)
+  if mixed
+    @variable(m, x>=0, Int)
+    @variable(m, y[1:2], Int)
+    @variable(m, s[1:5]>=0, Int)
+    @variable(m, l[1:5]>=0, Int)
+  else
+    @variable(m, x>=0)
+    @variable(m, y[1:2])
+    @variable(m, s[1:5]>=0)
+    @variable(m, l[1:5]>=0)
+  end
+  
+  @objective(m, Min, -x -3y[1] + 2y[2])
+
+  @constraint(m, -2x +  y[1] + 4y[2] + s[1] ==  16)
+  @constraint(m,  8x + 3y[1] - 2y[2] + s[2] ==  48)
+  @constraint(m, -2x +  y[1] - 3y[2] + s[3] == -12)
+  @constraint(m,       -y[1] + s[4] ==   0)
+  @constraint(m,        y[1] + s[5] ==   4)
+  @constraint(m, -1 + l[1] + 3l[2] +  l[3] - l[4] + l[5] == 0)
+  @constraint(m,     4l[2] - 2l[2] - 3l[3]               == 0)
+  for i in 1:5
+    @NLconstraint(m, l[i] * s[i] == 0)
+  end
+
+  JuMP.optimize!(m)
+
+  println("** Optimal objective function value = ", JuMP.objective_value(m))
+  println("** Optimal x = ", JuMP.value(x))
+  println("** Optimal y = ", JuMP.value.(y))
+  println("** Optimal s = ", JuMP.value.(s))
+  println("** Optimal l = ", JuMP.value.(l))
+  nothing;
+end
+
+model = direct_model(SCIP.Optimizer())
+solve_model(model)
+
+model = direct_model(SCIP.Optimizer())
+solve_model(model,mixed=true)
+
--- a/example/scip_with_jump_minlp.jl
+++ b/example/scip_with_jump_minlp.jl
@@ -1,3 +1,7 @@
+#=
+The following some part of code are excerpt from "Julia Programming for Operations Research 2/e (Changhyun Kwon) - Chap 8"
+=#
+
 using JuMP
 #using JuMP:MOI
 using SCIP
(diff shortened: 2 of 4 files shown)


Project.toml
```diff
@@ -8,7 +8,7 @@ Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 
 [compat]
-CEnum = "^0.1.0"
+CEnum = "^0.1.0, 0.2"
```
rschwarz: I'm not sure what this means. Did you require `CEnum` in version 0.2?
I don't think it's compatible with our generated wrapper code.

example/scip_with_jump_minlp.jl
```diff
@@ -0,0 +1,46 @@
...
+
+  @constraint(m, -2x +  y[1] + 4y[2] + s[1] ==  16)
+  @constraint(m,  8x + 3y[1] - 2y[2] + s[2] ==  48)
+  @constraint(m, -2x +  y[1] - 3y[2] + s[3] == -12)
+  @constraint(m,       -y[1] + s[4] ==   0)
+  @constraint(m,        y[1] + s[5] ==   4)
+  @constraint(m, -1 + l[1] + 3l[2] +  l[3] - l[4] + l[5] == 0)
+  @constraint(m,     4l[2] - 2l[2] - 3l[3]               == 0)
+  for i in 1:5
+    @NLconstraint(m, l[i] * s[i] == 0)
```
rschwarz: Since this is quadratic, it actually doesn't need the `@NLconstraint`, and could use `@constraint` instead. But it should not matter much to SCIP.

src/MOI_wrapper/results.jl
```diff
@@ -40,7 +40,7 @@ function MOI.get(o::Optimizer, ::MOI.ResultCount)::Int
 end
 
 function MOI.get(o::Optimizer, ::MOI.RawStatusString)
-    return String(CEnum.enum_name(SCIPgetStatus(o)))
+    return String(Symbol(SCIPgetStatus(o)))
```
rschwarz: This is probably a consequence of upgrading `CEnum`?

src/wrapper/commons.jl
```diff
@@ -700,7 +700,7 @@ const SCIP_PRIMAL = SCIP_Primal
 # Skipping MacroDefinition: SCIP_DECL_PROBEXITSOL ( x ) SCIP_RETCODE x ( SCIP * scip , SCIP_PROBDATA * probdata , SCIP_Bool restart )
 # Skipping MacroDefinition: SCIP_DECL_PROBCOPY ( x ) SCIP_RETCODE x ( SCIP * scip , SCIP * sourcescip , SCIP_PROBDATA * sourcedata , SCIP_HASHMAP * varmap , SCIP_HASHMAP * consmap , SCIP_PROBDATA * * targetdata , SCIP_Bool global , SCIP_RESULT * result )
 
-@cenum(SCIP_Objsense{Int32},
+@cenum(SCIP_Objsense,
```
rschwarz: All of this code is auto-generated and should not be edited.

Project.toml
mrchaos: It need version 0.2 or later to install SCIP interface version 7.4 above 
Current CEnum version is 0.2
New SCIP interface version not installed without 0.2

rschwarz: I don't understand. The newest version of SCIP.jl is 0.9.2 and the newest version of SCIP is 6.0.2. So what is this 7.4 refering to?

I did not try Julia 1.3.0 yet, so maybe it's related to that.

example/scip_with_jump_minlp.jl
mrchaos: Yes, I just wanted to know if @NLcontraint works well with SCIP.
The example code refers to Chap8 in "Julia Programming for Operations Research 2 / e".

src/MOI_wrapper/results.jl
mrchaos: The error occurs because CEnum does not have an enum_name field.

rschwarz: Well, it has in version 0.1.0: [CEnum.jl](https://github.com/JuliaInterop/CEnum.jl/blob/v0.1.0/src/CEnum.jl#L45)

example/scip_with_jump_minlp.jl
mlubin: There are potential copyright issues with copying code verbatim from a book. Please clarify the license of this code.

Project.toml
rschwarz: I just tried to `add SCIP` to a fresh Julia 1.3.0 installation, which worked just fine. It installed `CEnum 0.1.2` as a dependency and the tests passed. Maybe you have some other package that requires `CEnum 0.2`?

Also, see #144 for failing tests with `CEnum 0.2`. Your changes probably fix those.

So, while it does make sense to upgrade `CEnum` at some point, I would prefer to do it "properly", and regenerate the C wrappers with a new `Clang` etc.

example/scip_with_jump_minlp.jl
mrchaos: OK~ 
I'll check!

src/wrapper/commons.jl
mrchaos: SCIP_Objsense{Int32} type is "Expr" 

CEnum.jl
 77     elseif !isa(T, Symbol)  ==> ERROR
 78         throw(ArgumentError("invalid type expression for Cenum $T"))
 79     end

example/scip_with_jump_minlp.jl
chkwon: I'm not sure what copyright license my codes carry, but I'm good with this. 

Codes were mine, but the actual example is from: 

Ex 9.1.1 from Floudas, C.A., Pardalos, P.M., Adjiman, C., Esposito, W.R., Gümüs, Z.H., Harding, S.T., Klepeis, J.L., Meyer, C.A. and Schweiger, C.A., 2013. Handbook of Test Problems in Local and Global Optimization (Vol. 33). Springer Science & Business Media, http://titan.princeton.edu/TestProblems/

matbesancon:
there should probably be a separate PR for CEnum to keep things simple

rschwarz:
> there should probably be a separate PR for CEnum to keep things simple

This was already done and is merged now: #175.

So, I guess this PR could be rebased on master and simplified.
We could then also consider adding another CI test stage for testing examples using JuMP.

PR #149, opened 2020-04-01
Support SCIP 7.0.0

rschwarz:


rschwarz:
Still want to update Travis CI to use the newest SCIP, but the downloads are not ready yet?!

fserra:
i have no idea why it still doesn't find it. Can one do an "ls" on `C:\scipoptdir\lib`?

rschwarz:
It would be easiest to test this on a local Win machine first, I guess.

But yes, there is `Base.Filesystem.pwd()` and `Base.Filesystem.readdir()` in Julia. You can call `println( ... )` around these.

fserra:
scip now needs tbb. and in windows is not finding the tbb.dll so we need to install it i guess. Apparently it works out of the box in linux

rschwarz:
So, you mean that there is an additional dynamic library file, but it's missing from the installer?

This also implies that the error message from our `build.jl` is not really accurate, because in reality, the `scip.dll` file might exist, but `dlopen` fails because of some unresolved symbol or so?

fserra:
yes I guess. So the `libscip.dll` was ok (I changed it to libscip.lib which is wrong). But I guess it identifies that `libscip.dll`  is not really the dll needed because it doesn't resolve everything. And yes, the error is not really transparent in saying where the file was not found or not useful.

rschwarz:
@fserra: Windows now works with the new binaries, but OS X is failing when building from source, see https://travis-ci.org/github/SCIP-Interfaces/SCIP.jl/jobs/672463497#L208

By the way, I've previously tried to use pre-compiled binaries for OS X as well, but never got it to work, see #143.

fserra:
yeah, i have no idea. I opened an issue in scip.

rschwarz:
I tried again with preinstalled binaries of OS X, and it seems to be missing some dependencies:
https://travis-ci.org/github/SCIP-Interfaces/SCIP.jl/jobs/672477970#L218

PR #158, opened 2020-05-02
Implement MOI.DualStatus

### LINKED ISSUE #157: SCIP Dual Status
I got a report that SCIP.jl does not support the solution `dual_status` parameter.  I believe this parameter should be set to `NO_SOLUTION` if it is not available.

https://github.com/lanl-ansi/InfrastructureModels.jl/issues/68

CC @mlubin, @odow, @rb004f

odow:
I don't have SCIP installed locally, so didn't test. 

Closes #157

### DIFF
--- a/src/MOI_wrapper/results.jl
+++ b/src/MOI_wrapper/results.jl
@@ -31,6 +31,10 @@ function MOI.get(o::Optimizer, attr::MOI.PrimalStatus)
     end
 end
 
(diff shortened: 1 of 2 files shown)


rschwarz:
Should this be covered by some MOI test?

I'm not sure how to test this besides explicitly calling that function.

rschwarz:
OK, I see that you already added a test, thanks!

PR #164, opened 2020-05-15
add Aqua tests

rschwarz:


rschwarz:
Problem reported: https://github.com/JuliaLang/julia/issues/35909

matbesancon:
TIL you could have compat bounds on a package that is only in your tests

matbesancon:
no need anymore?

rschwarz:
I guess there was never a *need* per se, I just wanted to try this out on another package that depends on SCIP but that failed with an error.

I understand that there is a bug in Julia itself causing the error. This might now be fixed in `master`, but to apply it to SCIP, we would have to require Julia ≥ 1.6, and I find that to be too restrictive for the sake of adding some generic tests.

PR #167, opened 2020-05-24
undefined variable

matbesancon:
Found this while investigating the SOC business


rschwarz:
Thanks, looks like a copy-paste error. But I guess it should be `ConsRef`, not `VarRef`, as well?

rschwarz:
The failure on linux seems to come from broken SSL certifications on scip.zip.de.
Maybe this is a temporary issue? Or else we can adjust the `wget` command.

PR #172, opened 2020-07-20
Correct windows dll name

joaquimg:


joaquimg:
my bad, I was using v 6.0.2
the latest SCIP is indeed libscip.dll

Issue #174, opened 2020-07-26
Copy CEnum as submodule to prevent downgrading of other packages


GregPlowman:
I understand `SCIP` is not compatible with the latest `CEnum.jl`, and this is a known, outstanding issue.

Would it be feasible and would there be interest in copying the old required `CEnum` into the `SCIP` package as a submodule, and removing the dependency on the external `CEnum` package? (It seems this would be an easy change?)

Then the latest `SCIP` can co-exist with the latest `Clp`, `Cbc` in the same environment.

rschwarz:
Yes, I think this is feasible. Actually, in earlier versions of `Clang.jl`, a local `CEnum` module was created as part of the wrapper generation process, so there is precedent for it.

rschwarz:
fixed by #175

PR #175, opened 2020-07-27
RFC: Replace external dependency CEnum with private submodule

GregPlowman:
Attempt at implementing #174

Replace external dependency on older version of `CEnum` with a "local" submodule.
The aim is to prevent `SCIP` from holding back `CEnum`, which in turns holds back other packages (e.g. `Cbc`,  `Clp`).
In particular it would nice if the latest versions of `SCIP` and other solvers could coexist in the same environment.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,12 +3,10 @@ uuid = "82193955-e24f-5292-bf16-6f2c5261a85f"
 version = "0.9.5"
 
 [deps]
-CEnum = "fa961155-64e5-5f13-b03f-caf6b980ea82"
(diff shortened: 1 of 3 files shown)


rschwarz:
Thanks for the effort.

Do you think it makes sense to rename the `CEnum` submodule, in order to avoid conflict or confusion?

rschwarz:
I've tested installation of usage of both SCIP, Cbc and CEnum in a common environment and there seem to be no problems, so I will go ahead and merge & release.

PR #177, opened 2020-08-06
Add SCIP_jll as dependency

### LINKED ISSUE #165: BinaryBuilder package for libscip
Following [An update on SCIP](http://www.pokutta.com/blog/news/2020/05/15/scip-update.html), I tried putting together a [BinaryBuilder](https://binarybuilder.org/) package for the SCIP libraries, to make installation easier for Julia users.

Alas, my first attempt was not successful, so I will document what I tried so far:

Using `BinaryBuilder.run_wizard()`, I specified the source location [scipoptsuite-7.0.0.tgz](https://scip.zib.de/download/release/scipoptsuite-7.0.0.tgz) and also added several binary dependencies:

- `bliss`: for SCIP symmetry detection
- `boost`: for SoPlex multiprecision floats
- `Ipopt`: for SCIP's global (MI)NLP
- `GMP`: not sure if necessary, exact for ZIMPL and SoPlex rational arithmetic?

I started with the recommended `cmake` incantation given by the wizard. This detected all the expected dependencies, so I could continue with `make`.

----

Then I hit the first error: When building PAPILO, it will also try to build a vendored version of Intel TBB. But the configuration of TBB failed with: `getconf: GNU_LIBC_VERSION: unknown variable`.
Not sure if that is related to the build environment of BB based on an `alpine` Docker image, using `musl`?

I guess it would be best to provide a separate JLL package for Intel TBB, if PAPILO should be added.

----

In any case, I decided to disable some features and compile only the core library with SoPlex and SCIP. That is, I added the `cmake` flags `-DPAPILO=OFF -DZIMPL=OFF -DGCG=OFF`.

But now I got another error from SoPlex: Some static assert failed (forgot to copy), which makes me think that some more recent compiler version is required.

By default, BB uses GCC 4.8.5. It's possible to change the compiler, but I have not yet understood what that means for compatibility with other packages or Julia itself.

rschwarz:
Fix #165 .


rschwarz:
I might have introduced too many changes in the CI now, let's see...

rschwarz:
It looks OK now, except for this nasty error on AppVeyor for Windows:

```
Exception: EXCEPTION_ACCESS_VIOLATION at 0xffffffffc5160000 -- unknown function (ip: FFFFFFFFC5160000)
in expression starting at C:\projects\scip-jl\test\direct_library_calls.jl:3
unknown function (ip: FFFFFFFFC5160000)
Allocations: 5949752 (Pool: 5947818; Big: 1934); GC: 6
ERROR: Package SCIP errored during testing
Stacktrace:
 [1] pkgerror(::String) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Pkg\src\Types.jl:52
 [2] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, julia_args::Cmd, test_args::Cmd, test_fn::Nothing) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Pkg\src\Operations.jl:1566
 [3] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, test_fn::Nothing, julia_args::Cmd, test_args::Cmd, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Pkg\src\API.jl:328
 [4] #test#61 at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Pkg\src\API.jl:67 [inlined]
 [5] test(; name::Nothing, uuid::Nothing, version::Nothing, url::Nothing, rev::Nothing, path::Nothing, mode::Pkg.Types.PackageMode, subdir::Nothing, kwargs::Base.Iterators.Pairs{Symbol,Bool,Tuple{Symbol},NamedTuple{(:coverage,),Tuple{Bool}}}) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.5\Pkg\src\API.jl:80
 [6] top-level scope at none:1
```
I wouldn't know how to debug this, except for running more variations on AppVeyor.

rschwarz:
I tried reproducing the failing tests on Windows and succeeded: I get the same error message, but no additional information.
When I try to `include` the test file in a Julia REPL, it simply closes on me, burying the output.

When I run the lines of the first test one-by-one, I find that the first couple of `ccall`s actually work OK. It's only the call to `SCIPsolve` that crashes.

This makes me think that the compilation of SCIP itself might be fine, but that there is a problem with one of the linked libraries, such as SoPlex or Ipopt, which are only used at this point. Maybe the SoPlex library should be listed explicitly as a `LibraryProduct` after all? Then again, why does it work on all OSs?

matbesancon:
Does SCIP use the shared library or the executable? When building, make install only seems to install the SoPlex executable and not library

rschwarz:
Actually, I believe that the SoPlex code is already contained in `libscip.so`. Here's what I get for the `scip` executable in the `artifact` on my Linux:

```
# rs@t440p:~/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/bin [14:59:25]
$ ldd scip 
	linux-vdso.so.1 (0x00007ffda0d7c000)
	libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f78fa3ea000)
	libgmpxx.so.4 => /usr/lib/x86_64-linux-gnu/libgmpxx.so.4 (0x00007f78fa3e1000)
	libbliss.so => not found
	libipopt.so.3 => not found
	libquadmath.so.0 => /usr/lib/x86_64-linux-gnu/libquadmath.so.0 (0x00007f78fa39f000)
	libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f78fa21b000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f78fa096000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f78fa07c000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f78f9ebb000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f78fa4ac000)
```

What is surprising here, though, is the reference of `libgmp`. This is supposed to be disabled via the `cmake` flag.

matbesancon:
> libipopt.so.3 => not found

also surprising is the Ipopt not being there?

rschwarz:
No, I think that's fine. There is [Julia code](https://github.com/JuliaBinaryWrappers/SCIP_jll.jl/blob/master/src/wrappers/x86_64-linux-gnu-cxx11.jl#L38) in the `SCIP_jll` package that will take care of that.

It might be worth trying to add GMP back to the dependencies. Or else, apply more force to `cmake` to make sure it's not found during the build either.

But I'd rather find out which call is causing the error (within `SCIPsolve`). Not sure whether that would be visible if we compiled SCIP in debug mode and/or used `gdb` to run Julia.

matbesancon:
So I re-added GMP as a dependency on this branch: https://github.com/matbesancon/Yggdrasil/tree/scip-up
But I can't test for now, no access to a powerful enough machine

rschwarz:
So, I do have a notebook with Win 10 Pro on it, and could do some testing locally. Not sure if I will find the time before the holidays...

matbesancon:
I have access to a windows machine I'll be able to take a look

matbesancon:
I wonder if we should switch the URLs to scipopt.org from scip.zib.de, not sure if the older one will be kept

rschwarz:
Yes, that's a good idea.

I planned to make a separate PR to switch from Travis to GitHub actions and update some other minor things.

matbesancon:
Yes I also have to do it for other repos

README.md
```diff
@@ -8,6 +8,20 @@ Julia interface to [SCIP](http://scip.zib.de) solver.
...
 
+## Update (August 2020)
+
+It is no longer required to install [SCIP](http://scip.zib.de) itself before you
+can use this package. There now exists a
+[BinaryBuilder.jl](https://github.com/JuliaPackaging/BinaryBuilder.jl) generated
+package [SCIP_jll.jl](https://github.com/JuliaBinaryWrappers/SCIP_jll.jl) which
+is installed automatically as a dependency.
+
+So, with if you use Julia version 1.3 or newer, you can get started simply with
```
GregPlowman, line 19: “So if you use ...”

matbesancon:
So I had forgotten how awful all the tooling can be on windows. For some reason the error message I get is in French:
image
That happens on `]add SCIP_jll`

matbesancon:
And when testing SCIP with the branch:
image

matbesancon:
the error message is awfully unhelpful here

matbesancon:
The first one could just be a network error

matbesancon:
@Giordano is the last error message issued from a JLL package?

giordano:
Do you mean
```
julia: malloc.c:4033: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```
?

matbesancon:
Sorry no the message in screenshot from the SCIP test, with
"Please submit a bugf report with steps to reproduce this fault"

giordano:
Maybe?  No idea

matbesancon:
ok thanks, I wasn't sure if it wasn't a default error message for JLL packages

giordano:
Oh, no no, they shouldn't throw any error at all (maybe only if someone tries to use them in old Julia versions)

matbesancon:
@rschwarz in addition to the windows issue, do you know what is causing the travis fail?

rschwarz:
Oh my, that all looks worse than I remember ;-)

So, your French error message seems related to downloading artifacts and some method signature has changed?

The second one looks the same as to the [failure we had in August on AppVeyor](https://ci.appveyor.com/project/rschwarz/scip-jl/builds/34531572/job/r2hcagfhg9gnuwwb):
```
Please submit a bug report with steps to reproduce this fault, and any error messages that follow (in their entirety). Thanks.
Exception: EXCEPTION_ACCESS_VIOLATION at 0xffffffffd82e0000 -- unknown function (ip: FFFFFFFFD82E0000)
in expression starting at C:\projects\scip-jl\test\direct_library_calls.jl:3
unknown function (ip: FFFFFFFFD82E0000)
Allocations: 16069118 (Pool: 16065129; Big: 3989); GC: 14
ERROR: Package SCIP errored during testing
Stacktrace:
 [1] pkgerror(::String, ::Vararg{String,N} where N) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Pkg\src\Types.jl:53
 [2] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, julia_args::Cmd, test_args::Cmd, test_fn::Nothing) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Pkg\src\Operations.jl:1510
 [3] test(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; coverage::Bool, test_fn::Nothing, julia_args::Cmd, test_args::Cmd, kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Pkg\src\API.jl:316
 [4] #test#68 at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Pkg\src\API.jl:297 [inlined]
 [5] #test#65 at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\Pkg\src\API.jl:294 [inlined]
 [6] top-level scope at none:1
```

But, back then, the tests were still passing in Travis. The [latest failure](https://travis-ci.org/github/scipopt/SCIP.jl/builds/747293569) only occurs on Julia 1.5, but maybe back then, it was tested with 1.4?

Looking at the output of the failing test (in the `pr`, not `push), it looks a little different:
```
corrupted size vs. prev_size
```
versus
```
julia: malloc.c:4033: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.
```
And these are failures on Linux with Julia 1.5, so it would be best to reproduce them locally, without CI.

Also, when I pushed that other commit (to upgrade to 7.0.1) that I found in my local branch, it seems that AppVeyor is not even triggered anymore?

Other than that, it might be worth to first do the move to GitHub actions (does it support Windows?) and then come back to this PR.

matbesancon:
So the two errors in French look like 404 errors when attempting to download the artifacts. The fact that it's in French make me suspect git more than the Julia tooling

matbesancon:
Ok there is definitely a bug not only on travis. The first test where it happens is
```
    MOIT.qcptest(BRIDGED, CONFIG)
```

It is the first test requiring Ipopt I think, so it might be that

matbesancon:
Same error happens on:
```
julia> MOIT.qptest(BRIDGED, CONFIG)
julia: malloc.c:4033: _int_malloc: Assertion `(unsigned long) (size) >= (unsigned long) (nb)' failed.
```

matbesancon:
Ok digging deeper, I had an error that may be more useful. It happens in the `MOIT.qp1test` when calling optimize!
```
julia> MOI.optimize!(model)

******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt
******************************************************************************

munmap_chunk(): invalid pointer

signal (6): Aborted
in expression starting at REPL[44]:1
gsignal at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
abort at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
unknown function (ip: 0x7f66a8229966)
unknown function (ip: 0x7f66a82309d9)
cfree at /lib/x86_64-linux-gnu/libc.so.6 (unknown line)
dmumps_f77_ at /home/mbesancon/.julia/artifacts/ad5ff2a3f16cc91386d488d897a8d9a36cb5e849/lib/libdmumps.so (unknown line)
dmumps_c at /home/mbesancon/.julia/artifacts/ad5ff2a3f16cc91386d488d897a8d9a36cb5e849/lib/libdmumps.so (unknown line)
_ZN5Ipopt20MumpsSolverInterfaceD2Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt20MumpsSolverInterfaceD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt16TSymLinearSolverD2Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt16TSymLinearSolverD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt18StdAugSystemSolverD1Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt18StdAugSystemSolverD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt17PDFullSpaceSolverD2Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt17PDFullSpaceSolverD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt21PDSearchDirCalculatorD2Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt21PDSearchDirCalculatorD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt14IpoptAlgorithmD1Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt14IpoptAlgorithmD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt16IpoptApplicationD2Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZN5Ipopt16IpoptApplicationD0Ev at /home/mbesancon/.julia/artifacts/8cd817851eb05b18fded2decd4f3e67abe1e983d/lib/libipopt.so (unknown line)
_ZL20nlpiFreeProblemIpoptP9SCIP_NlpiPP16SCIP_NlpiProblem at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPnlpFree at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPfreeSolve at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPfreeTransform at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
solveSubNLP at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPapplyHeurSubNlp at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
heurExecSubNlp at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPheurExec at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPprimalHeuristics.part.12 at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
solveNodeLP.isra.14 at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
solveNode.isra.16 at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPsolveCIP at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPsolve at /home/mbesancon/.julia/artifacts/d94fb1efe141bd5787829453f44129800a77305d/lib/libscip.so (unknown line)
SCIPsolve at /home/mbesancon/Documents/crptests/project_julia/SCIP.jl/src/wrapper/scip_solve.jl:14 [inlined]
optimize! at /home/mbesancon/Documents/crptests/project_julia/SCIP.jl/src/MOI_wrapper.jl:223 [inlined]
optimize! at /home/mbesancon/.julia/packages/MathOptInterface/ZJFKw/src/Bridges/bridge_optimizer.jl:264
```

matbesancon:
So the problem is a function called in MUMPS from Ipopt that does not exist?

rschwarz:
Maybe? `munmap_chunk(): invalid pointer` could also mean that some memory is corrupted.

Not sure if MUMPS & IPOPT could be built in some kind of debug-mode that would fail earlier with some assertion.
But IPOPT is also pulled in from a `_jll`, right?

I'll reiterate my modest suggestion of first providing a SCIP_jll without IPOPT.
I feel like this would work well for most casual users of SCIP.jl.
The expert users might want to build their own SCIP (or use a `deb` package etc) in any case, because they want a better linear solver for IPOPT which can't be distributed within the IPOPT_jll.

matbesancon:
Ok yes, I guess we can start here, I'll modify the SCIP_jll build

matbesancon:
do we need a `-DIPOPT=0` in the cmake?

rschwarz:
> do we need a `-DIPOPT=0` in the cmake?

That probably helps, but it should automatically disable it when not found (that is, when you drop the dependency).

matbesancon:
I bumped the patch version in SCIP_jll to v0.1.1, so we will need to use the new one

rschwarz:
OK, but only after it's merged and published, right?

matbesancon:
Yes, it's just been merged, the SCIP_jll should be out soon

rschwarz:
Works locally for me.

rschwarz:
Still not sure why AppVeyor is not triggered. Maybe because the GitHub organization was renamed?

giordano:
You may want to simply move to GitHub Actions, as you're probably going to run out of Travis credits (unless you're paying for it or you've got the free OSS credits)

matbesancon:
Yes I think that's the next step

matbesancon:
@rschwarz yes it seems the appveyor repo needs to be updated, in "settings", update the github repo to scipopt/SCIP.jl

matbesancon:
dumb question, how come the quadratic tests pass even though we don't have Ipopt anymore?

rschwarz:
It seems that there are still failures because of incompatible GMP versions. This is surprising because GMP should not be used at all by SCIP_jll. I guess the `cmake` incantation is still not explicit enough.

Also, SCIP does not require IPOPT for solving nonlinear problems, because the main approach is based on spatial branching. It's "only" used for heuristics, I believe.

matbesancon:
I'm not sure which GMP version this is referring to.
https://travis-ci.org/github/scipopt/SCIP.jl/jobs/747401103#L280
GMP goes up to 6.2, does the dynamic lib have different version numbers?

giordano:
The compilation log shows
```
[10:24:59] -- The following OPTIONAL packages have been found:
[10:24:59] 
[10:24:59]  * Quadmath
[10:24:59]  * GMP
[10:24:59]  * BLISS
```

matbesancon:
this is weird, we have a `-DGMP=OFF` flag
https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SCIP/build_tarballs.jl#L19

matbesancon:
Should we build using the shared library only? From the SCIP docs:
```
## Creating a single library containing SCIP, SoPlex, and ZIMPL

In case you need a single library which contains SCIP, SoPlex, and ZIMPL
you can use the command:

    make scipoptlib

This will create a single library containing SCIP, SoPlex, and ZIMPL. It is
placed in the `lib` directory. This library is statically linked. If you
need a shared library use the command:

    make scipoptlib SHARED=true
```

matbesancon:
Weird, it is currently working for me in a docker image (without SCIP installed before `]add SCIP`)

rschwarz:
I'm not sure why `-DGMP=OFF` is not working. Maybe it's overriden by some other part (CMAKE is used recursively). But most optional things are disabled already (GCG, Zimpl, ...). I also don't remember why we sometimes use `OFF` or `off` or `0` to disable features here.

Using `make scipoptlib` sounds like a good idea, because it would save time and disk space. However, I vaguely remember a related change from SCIP 6 to SCIP 7. I think the name of the library changed from `libscipopt.so` to `libscip.so` but also the latter now contains SoPlex etc by default.

Re: your last point, I'm not sure what you mean by 'it works'. I guess that the package build should fail when SCIP is not found.

giordano:
> Weird, it is currently working for me in a docker image (without SCIP installed before ]add SCIP)

That's a problem only with GMP on macOS.

rschwarz:
I now believe that the `bliss` dependency brings GMP back in. I tried (locally) to disable everything optional, and the CMAKE output no longer shows GMP in the FOUND OPTIONAL... section.

Then I installed `libbliss-dev` on my Debian system and ran CMAKE again, but now it found GMP and included it.

So, we (might) have to

- add GMP as a dependency to SCIP_jll, or 
- remove bliss as a dependency from SCIP_jll

giordano:
I think the fundamental problem is that the CMake script uses GMP when it's found and it doesn't check if it was actually requested, but GMP may have been found for some other dependencies

rschwarz:
Makes sense, so the safe choice is to include GMP as a dependency at the jll level.

giordano:
I believe
```cmake
# GMP
if(GMP)
    find_package(GMP)
endif()
if(GMP_FOUND)
    include_directories(${GMP_INCLUDE_DIRS})
    set(SCIP_WITH_GMP on)
else()
    set(GMP_LIBRARIES "")
endif()
```
should be something like
```cmake
# GMP
if(GMP)
    find_package(GMP)
    if(GMP_FOUND)
        include_directories(${GMP_INCLUDE_DIRS})
        set(SCIP_WITH_GMP on)
    else(GMP_FOUND)
        set(GMP_LIBRARIES "")
    endif(GMP_FOUND)
endif(GMP)
```

rschwarz:
I've prepared the branch [rs/github_action](https://github.com/scipopt/SCIP.jl/tree/rs/github_action) and PR #180 which depends on SCIP_jll working.

Maybe that branch should be merged into this one, once SCIP_jll is updated w.r.t. GMP?

rschwarz:
I started updating the SCIP_jll build_tarballs.jl, but I want to first test it locally. Turns out that I get errors about the `Platform` constructor with the most recent BinaryBuilder, so I'll use `supported_platforms()`.

rschwarz:
> Using `make scipoptlib` sounds like a good idea, because it would save time and disk space. 

I looked up this part of the documentation as well. I think this Makefile target only exists in the old-style build system of SCIP, but not in the new one (based on CMAKE).

matbesancon:
OK, this was maybe palliating the lack of flexibility of the make system

matbesancon:
Could this have an impact?
https://scipopt.org/doc/html/FAQ.php#scipblisscrash

rschwarz:
I don't know about this bliss crash. The answer says that CMAKE should detect it :man_shrugging:

rschwarz:
I'm not sure whether the most recent CI runs (now with GH action) already use the most recent SCIP_jll, but:
 - Linux works fine
 - OS X fails with GMP incompatibility
 - Windows fails with the same errors as before (so it's not directly related to Ipopt, it seems)

matbesancon:
OSX seems fine now

rschwarz:
OK, so we're back where we started :roll_eyes:

matbesancon:
well at least we are less breaking :sweat_smile:

rschwarz:
In August, Linux and OS X were also working, but yeah...

I guess one avenue for further testing would be to build SCIP_jll in debug mode and see whether we get more detail / context for the failure in Windows. I'd prefer to do that locally, without the delay of PRs and package registrations.

matbesancon:
So digging into the windows issue, it looks like building the problem (in direct and managed mode) does not yield to an error, but calling SCIPsolve does

matbesancon:
I built the SCIP executable with BB along with the library.
Running from the SCIP CLI, things look ok up to the optimize point:

```


optimize
SCIP>
LP Solver <SoPlex 5.0.1>: barrier convergence tolerance cannot be set -- tolerance of SCIP and LP solver may differ
LP Solver <SoPlex 5.0.1>: fastmip setting not available -- SCIP parameter has no effect
LP Solver <SoPlex 5.0.1>: number of threads settings not available -- SCIP parameter has no effect
transformed problem has 3 variables (0 bin, 0 int, 0 impl, 3 cont) and 3 constraints
      3 constraints of type <linear>

original problem has 6 active (66.6667%) nonzeros and 6 (66.6667%) check nonzeros
```

Could it be that the LP solver is not found / not linked?

rschwarz:
I think these 3 warnings can be explained by SCIP setting some LP-related parameters that not all LP solvers support. For example, SoPlex is sequential, so setting the number of threads is meaningless.

It would stil be strange if these messages only appeared in Windows.

What happens afterwards? Does the process just quit, or does it actually solve the problem?

matbesancon:
It hangs forever, it does not respond anymore and need a ^C signal to quit
(it does not respond to commands after "optimize")

rschwarz:
OK, strange, so it's failing already when running the binary, without involving Julia in any way?
That would make it easier to report a bug. It would be interesting to reproduce that without BinaryBuilder, just calling cmake and make. Maybe the specific compiler version is important to trigger the problems?

matbesancon:
> OK, strange, so it's failing already when running the binary, without involving Julia in any way?

yes, the only involvement is that the binaries were built using BB and obtained by downloading the SCIP_JLL lib

rschwarz:
I guess that was too early :disappointed: I thought I had seen a notification about a new registered version :confused:

giordano:
https://github.com/JuliaPackaging/Yggdrasil/pull/2288#discussion_r546260256

rschwarz:
Unfortunately, [no change in behavior](https://github.com/scipopt/SCIP.jl/pull/177/checks?check_run_id=1582746776#step:6:65)

.appveyor.yml
```diff
@@ -1,30 +0,0 @@
...
-
-notifications:
-  - provider: Email
-    on_build_success: false
-    on_build_failure: false
-    on_build_status_changed: false
-
-install:
-  - ps: iex ((new-object net.webclient).DownloadString("https://raw.githubusercontent.com/JuliaCI/Appveyor.jl/version-1/bin/install.ps1"))
-  - ps: wget https://scip.zib.de/download/release/SCIPOptSuite-7.0.0-win64-VS15.exe -outfile scipopt-installer.exe
```
matbesancon, line 21: Should we re-add appveyor with the download updated to latest patch?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ['1']  # Test against current minor release (at least 1.3)
```
matbesancon, line 14:
```suggestion
        version: ['1.3', '1']  # Test against current minor release (at least 1.3)
```
This would allow to test against the minimal version (1.3) and latest stable

Project.toml
```diff
@@ -5,9 +5,11 @@ version = "0.9.7"
 [deps]
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
+SCIP_jll = "e5ac4fe4-a920-5659-9bf8-f9f73e9e79ce"
 
 [compat]
 MathOptInterface = "^0.9.6"
+SCIP_jll = "^0.1.2"
 julia = "^1.0.0"
```
matbesancon, line 13:
```suggestion
julia = "1"
```
is simpler and equivalent

.appveyor.yml
rschwarz: Well, maybe. this change was merged from another branch where I wanted to migrate to GitHub actions.
Re-adding the AppVeyor CI would only make sense to me if we give up on SCIP_jll for Windows, I guess.

matbesancon:
well that's new

rschwarz:
Oh my :cry:

Well, the stacktrace of the failure on Ubuntu with Julia 1.3 is deep within IPOPT. And IPOPT wasn't included in SCIP_jll before, right?

matbesancon:
Yes exactly

matbesancon:
Could it be that SCIP requires a specific version of Ipopt that is not compatible with the one we have through Ipopt_jll?

matbesancon:
In the tests, this happens for me on 1.5 with the following:
```
@testset "MOI Quadratic Constraint" begin
    # needs objective bridge (MOI/#529)
    # MOIT.qptest(BRIDGED, CONFIG)

--->     MOIT.qcptest(BRIDGED, CONFIG)
end

julia> MOIT.qcptest(BRIDGED, CONFIG)

******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt
******************************************************************************

free(): invalid next size (normal)

signal (6): Aborted
in expression starting at REPL[9]:1
```

rschwarz:
> Could it be that SCIP requires a specific version of Ipopt that is not compatible with the one we have through Ipopt_jll?

Not that I know of. I used to simply compile IPOPT from the tarball available from COIN-OR (latest version).

matbesancon:
The weird part is that Ipopt seems to be loaded? I can't figure out at which point this is triggered

matbesancon:
@odow have you experienced something similar with Ipopt? We used the JLL here

rschwarz:
Yeah, it's not that IPOPT is misisng, but it tries to make some kind of syscalls(?) that are not supported. Maybe the IPOPT_jll and SCIP_jll are compiled in some incompatible way?

matbesancon:
Just cross-posted to slack. The biggest smell right now is that the bug only occurs on linux

matbesancon:
OK there was something strange with Ipopt 3.13.1 breaking, this is fixed on 3.13.2

matbesancon:
If it doesn't pick it up, I added a dependency of SCIP.jl on Ipopt_jll with an explicit requirement

matbesancon:
Windows still failing "as expected", but we now have Ipopt integration

rschwarz:
OK, great. I'll try to build in the Windows exception in the `build.jl` some time soon now.

odow:
There was an issue with MUMPS: https://github.com/jump-dev/Ipopt.jl/issues/246

matbesancon:
> Edit: no idea on the Windows side.

the Windows issue is on SCIP, so Windows users will have to provide a SCIP install, while unix users can get the package shipped with a default SCIP (they can but don't have to provide one)

matbesancon:
> There was an issue with MUMPS: jump-dev/Ipopt.jl#246

it might have been that, combined with the fact that the package manager had not picked the latest Ipopt_jll for reasons I didn't get

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    types: [opened, synchronize, reopened]
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ['1.3', '1']  # Test against current minor release (at least 1.3)
+        os: [ubuntu-latest, macOS-latest, windows-latest]
```
matbesancon, line 15:
```suggestion
        os: [ubuntu-latest, macOS-latest]
```

remove windows for now

README.md
```diff
@@ -2,12 +2,26 @@
...
 
 See [NEWS.md](https://github.com/SCIP-Interfaces/SCIP.jl/blob/master/NEWS.md) for changes in each (recent) release.
 
+## Update (August 2020)
+
+It is no longer required to install [SCIP](http://scip.zib.de) itself before you
+can use this package. There now exists a
+[BinaryBuilder.jl](https://github.com/JuliaPackaging/BinaryBuilder.jl) generated
+package [SCIP_jll.jl](https://github.com/JuliaBinaryWrappers/SCIP_jll.jl) which
+is installed automatically as a dependency.
```
matbesancon, line 17:
```suggestion
On MacOS and Linux, it is no longer required to install the [SCIP](https://scipopt.org/) binaries using this package. There now exists a
[BinaryBuilder.jl](https://github.com/JuliaPackaging/BinaryBuilder.jl) generated
package [SCIP_jll.jl](https://github.com/JuliaBinaryWrappers/SCIP_jll.jl) which
is installed automatically as a dependency.


On Windows, the separate installation of SCIP is still mandatory.
```

README.md
```diff
@@ -2,12 +2,26 @@
...
+
+So, with if you use Julia version 1.3 or newer, you can get started simply with
+
+    pkg> add SCIP
+
+See below for support of custom SCIP installations.
```
matbesancon, line 23:
```suggestion
Under Julia 1.3 or more recent, you can use this default installation:

    pkg> add SCIP

If you use an older Julia version, Windows or want a custom SCIP installation, see below for the build steps.
```

rschwarz:
I wonder whether `deps/build.jl` and `src/init.jl` should still be modified to print some warning for Windows users.

Issue #156, opened 2020-04-25
MOI callbacks


matbesancon:
Now that callbacks have been shipped in MOI, they can be implemented with the constraint handler system. A good example is GLPK or CPLEX: https://github.com/JuliaOpt/CPLEX.jl/blob/master/src/MOI/MOI_callbacks.jl

rschwarz:
True. Since SCIP constraint handlers need to implement 3-4 different callbacks (see [comment](https://github.com/SCIP-Interfaces/SCIP.jl/issues/94#issuecomment-469264567) in #94), there is not a faithful representation of them as a single lazy constraint callback.

One approximation would be to
- always call the lazy constraint callback (user defined) for any SCIP callback,
- use the generated constraint only in the `enforce` callbacks,
- implement `check` by counting the generated constraints (none, or more),
- implement `lock` by adding locks for all variables in both directions.

This is what we did in MPB/CSIP, but it's somewhat pessimistic (too many variable locks) and maybe confusing (`check` callback is called often, but generated cuts never added to problem).

matbesancon:
From what I see from the issue, the only point of ambiguity is on a lock strategy? As I understand them, the only variables for which lock is true would be the ones used either in the constraint or in the condition to check the constraint?

rschwarz:
Yes, the constraint handler must add locks to all variables where changing the assignment (e.g. rounding) to a lower or higher value might violate the constraint.

matbesancon:
To come back to this issue, three possibilities I see:
- Offer a vanilla type of callback (with all variables of the model)
- use something like IRTools to analyze the callback function when it is passed and extract the variables it touches
- when implementing the [callback set function](https://github.com/jump-dev/CPLEX.jl/blob/master/src/MOI/MOI_callbacks.jl#L12), add a last argument `variables` with default to `allvariables(optimizer)`, so that users can tune this default.

rschwarz:
I like the last option best, because it's still convenient for the user, but they are aware that there is a decision to be made.

matbesancon:
I'm not sure where option 2 would be located in terms of compromises and advantages, do you know of anyone who has used it for similar use cases?
Pinging @odow to know if some other options have been considered for other solvers

rschwarz:
I have not used `IRTools` knowingly. Analyzing the user code sounds a bit shaky to me, as well as difficult to implement.

odow:
[...]

PR #179, opened 2020-11-28
Implement Separators and MOI.UserCutCallback

jovarga:
There is still a problem with the tests: If there are two optimal solutions and one of them is cutted in a separator (or the cutcallback), SCIP still returns that solution, if it would have returned that solution in the first place. See also #156.


rschwarz:
At a first glance, the code looks good to me.

I have two questions for understanding:
 - There can only be at most one cut callback for any `MOI.Optimizer` right?
 - When users `MOI.set` the callback, they can't set any of the keyword arguments (priority, frequency, etc.) but will use defaults, right?
 - If the priority is fixed for MOI users, should it be set higher that 0, so that the user's callback is called before SCIP's generic separators? Not sure what the expectation is in other solvers.

For the tests, is the issue that the solution is not unique and you don't know which one to cut off?
Could this not be helped by changing the coefficients and make it non-symmetric (e.g. `2x + 3y ≤ 4`)?

Would it also make sense to add a valid cut that would separate the LP relaxation (e.g. `x + y ≤ 1`) and disable all of SCIP's separation and branch-and-bound, to see whether it would then be solved correctly?

From my point of view, it's also OK to add an *invalid* cut (which is too strong) just for the tests. So, you could add `x + y ≤ 0` and verify that the returned solution is worse than the actual optimum. This would only work if the priority of the separator is high enough, I guess, and SCIP doesn't find a solution in any other way :worried:

jovarga:
> - There can only be at most one cut callback for any `MOI.Optimizer` right?

Right. At least as far as I know.

> - When users `MOI.set` the callback, they can't set any of the keyword arguments (priority, frequency, etc.) but will use defaults, right?

Right.

> - If the priority is fixed for MOI users, should it be set higher that 0, so that the user's callback is called before SCIP's generic separators? Not sure what the expectation is in other solvers.

I'm not sure either. But I don't think there's a problem with raising the priority. Would make more sense to me.


Maybe I'll give some more details about the problem, that I have with the tests. There are two tests, using the same MIP:
```
max x + y
s.t. x + y <= 1.5
x,y in {0,1}
```
This problem has the two different solutions `(x,y)=(1,0)` and `(x,y)=(0,1)`. One of the tests cuts one of these solutions (with `x <= 0`), the other cuts the other solution (`y <= 0`). Still the solution, that is found by SCIP is `(x,y)=(1,0)` for both tests. Also the debug output shows, that no cuts have been added and that the number of rows stays at 1. Maybe there is a parameter, that prohibits the cut to enter the LP, such as `separating/minefficacy` or `separating/minortho`?

> For the tests, is the issue that the solution is not unique and you don't know which one to cut off?

That the solution is not unique is on purpose to not violate SCIP's expectations to the added cut as we would have undefined behavior otherwise.

> Would it also make sense to add a valid cut that would separate the LP relaxation (e.g. `x + y ≤ 1`) and disable all of SCIP's separation and branch-and-bound, to see whether it would then be solved correctly?

Do you know, which parameters I have to set to achieve that? I only found the [list of all SCIP parameters](https://www.scipopt.org/doc/html/PARAMETERS.php), but there are quite some...

> From my point of view, it's also OK to add an invalid cut (which is too strong) just for the tests. So, you could add x + y ≤ 0 and verify that the returned solution is worse than the actual optimum. This would only work if the priority of the separator is high enough, I guess, and SCIP doesn't find a solution in any other way

How to make sure, that SCIP doesn't find a solution in any other way?

rschwarz:
> Maybe there is a parameter, that prohibits the cut to enter the LP, such as separating/minefficacy or separating/minortho?

That's possible. Another explanation is that a solution was found by SCIP before the LP is solved or the separators are called (e.g. by a simple heuristic).

> Do you know, which parameters I have to set to achieve that?

I don't know for sure, but a good strategy is to set an `emphasis` and then look at what parameters SCIP sets as a result. For example:
```
SCIP> set sepa emph off
separating/disjunctive/freq = -1
separating/impliedbounds/freq = -1
separating/gomory/freq = -1
separating/strongcg/freq = -1
[...]
```

Similarly for `heuristics`. Another one would be `limits/nodes = 0`.

jovarga:
I additionally set the parameters from `set sepa emph off` and `set heuristics emph off` and now the tests work! I also added a test with an invalid cut.

I'll add some comments and from my point of view it is ready to be reviewed then.

rschwarz:
It seems that the tests still fail (consistently?) on CI. Do they all pass for you locally now? I can reproduce the failure on my system:
```
AddSingleCut (cut off another optimal solution): Test Failed at /home/rs/.julia/dev/SCIP/test/sepa.jl:126
  Expression: ≈(MOI.get(optimizer, MOI.ObjectiveValue()), 1.0, atol = atol, rtol = rtol)
   Evaluated: 0.0 ≈ 1.0 (atol=1.0e-6, rtol=1.0e-6)
Stacktrace:
 [1] top-level scope at /home/rs/.julia/dev/SCIP/test/sepa.jl:126
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/rs/.julia/dev/SCIP/test/sepa.jl:90
AddSingleCut (cut off another optimal solution): Test Failed at /home/rs/.julia/dev/SCIP/test/sepa.jl:127
  Expression: ≈(MOI.get(optimizer, MOI.VariablePrimal(), x), 1.0, atol = atol, rtol = rtol)
   Evaluated: 0.0 ≈ 1.0 (atol=1.0e-6, rtol=1.0e-6)
Stacktrace:
 [1] top-level scope at /home/rs/.julia/dev/SCIP/test/sepa.jl:127
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/rs/.julia/dev/SCIP/test/sepa.jl:90
Test Summary:                                      | Pass  Fail  Total
separators                                         |   14     2     16
  dummy sepa (no separation)                       |    1            1
  AddSingleCut sepa (cut off one optimal solution) |    5            5
  AddSingleCut (cut off another optimal solution)  |    3     2      5
  AddSingleCut (too strong cut)                    |    5            5
```

rschwarz:
I think that symmetry detection and `cons_orbitope` is to blame for this behavior.
Could you add the following parameter as well? 

```diff
+    setter("misc/usesymmetry", 0)
```

jovarga:
I just added that parameter as well.

test/cutcallback.jl
```diff
@@ -0,0 +1,119 @@
+using MathOptInterface
+const MOI = MathOptInterface
+using JuMP
```
rschwarz, line 3: This is not needed and should be removed.

All relevant types & methods are defined in MOI already, so JuMP is only for the end users.
Also, this makes the tests fail now, because JuMP is not even installed.

I wonder why this didn't fail before.

rschwarz: I only now noticed that JuMP is used to create the models for the tests :see_no_evil:.
Solvers should not depend on JuMP, so I'm afraid that the models have to be created "by hand" using MOI.

jovarga:
All four jobs on CI fail now with the following message:
```
Got exception outside of a @test
LoadError: ArgumentError: Package JuMP not found in current path:
- Run `import Pkg; Pkg.add("JuMP")` to install the JuMP package.
```
How to resolve that, i.e. how should JuMP be installed on CI? Alternatively I can also rewrite the tests not to use JuMP. However I would prefer the first solution.

rschwarz (changes requested):
Yes, I saw the failure as well, sorry that I overlooked that before.
Depending on JuMP is not good for solvers, so an MOI-generated model is much preferred.

test/cutcallback.jl
jovarga: > I wonder why this didn't fail before.

Probably because one of the tests failed before. There are two files for the tests and only one of them uses JuMP.

I'll replace the JuMP calls with MOI calls then.

jovarga: I'm finished with the replacement now. On my machine all testcases pass. Let's see if they also do on CI.

rschwarz: Thanks, looking good so far.

rschwarz (approved):

rschwarz:
Thanks for your efforts!

I assume that you'd rather use this functionality soon, so I'll try and make a new relese now.
Note that it usually takes a few days for these to be registered, because they need to be approved manually.

jovarga:
> I assume that you'd rather use this functionality soon, so I'll try and make a new relese now.

Thank you  :)

jovarga:
It seems the release has not been approved: [JuliaRegistries SCIP v0.9.7](https://github.com/JuliaRegistries/General/pull/25642)

rschwarz:
Yeah, it still needed manual approval, because the `] build SCIP` does not work in CI.
This will all go smoother once we get the `SCIP_jll` running.

rschwarz:
The latter, see also [General](https://github.com/JuliaRegistries/General#registering-a-package-in-general).

jovarga:
So someone of the Julia Team will take care of it. Just to have a rough estimate: How long did it take for past pull requests to be approved?

rschwarz:
I think it was usually merged after 2-3 days, so I don't know why it timed out now.

PR #182, opened 2021-01-07
Some tests avoided on < 7.0

matbesancon:


rschwarz (approved):
OK, now I understand your comment: You want to relax the upper bound.
This is fine with me, although I don't think that SCIP follows semver strictly.

Issue #77, opened 2019-01-04
Automate the wrapper generation, upgrade CEnum dep.


rschwarz:
This applies only after #76 is merged.

We now use Clang.jl to generate Julia wrappers from the SCIP header files automatically.
There is a script for that purpose that should be rerun for every (major) release of SCIP that we want to support.

Unfortunately, the process was not completely automatic. The steps should be documented clearly, or integrated properly in the generator script.

In particular, the following steps followed the execution of the script:
- rename `Ptr{SCIP}` to `Ptr{SCIP_}`, to avoid name conflict with module.
- delete empty wrappers for `type_*.h`; the type definitions all went to `commons.jl`
- add some missing (type/enum) definitions to `manual_commons.jl` (why did Clang.jl not add them?)

Also, we used the branch of Clang.jl from PR [#210](https://github.com/ihnorton/Clang.jl/pull/210), which was now merged to `master`, but not yet tagged?

leethargo:
Actually, the new Clang.jl seems to be tagged now (at 0.8).

Other potential improvements:
* Automatically generate the `wrapper.jl` file (with all the `include()` statements).
* Copy documentation strings from SCIP's headers to our Julia functions.
* Use `Ref{Ptr{}}` instead of `Ptr{Ptr{}}`?
* Use a stricter function signature for our `ccall` wrappers.

rschwarz:
This is also based on `Clang.jl`, so it would be mostly for convenience of configuration?

rschwarz:
OK, good to know.

Well, the variadic functions that are used do actually work now, there is just a risk of segfault or other unexpected behavior when targeting a new platform (or C compiler?).

I would have a look in the future, in case a re-generation of the wrapper becomes necessary after a release of a new SCIP version or similar. Right now, it works well enough for me not to bother, but PRs are welcome!

rschwarz:
Another topic that has come up on [Discourse](https://discourse.julialang.org/t/finalizer-only-works-with-mutable-structs/28777/9?u=leethargo). The auto-generated `ccall` wrappers all use `Ptr{X}` for some SCIP struct `X`. This is fine in most cases, as we will actually pass C pointers (for SCIP-owned objects) to these calls.

[...]

PR #186, opened 2021-03-03
Upgrade wrapper generator script and regenerate bindings

Gnimuc:
I'm testing https://github.com/JuliaInterop/Clang.jl/pull/278. Before making a new release for Clang.jl, I'd like to make sure that the code generated by the new generator is not broken. Any feedback about the new generator would be highly appreciated.


rschwarz:
Note that the generation process was not totally automatic in our case, see #77 .

We also pinned `CEnum` to an older version (for no good reason except that the we didn't want to change the generated code) and later *vendored* the module.

Gnimuc:
With this PR, the generation process should be totally automatic. 

Missing definitions are all macro-related and one can manually add them to `prologue.jl`. 

Variadic functions can be supported by using Julia 1.5's `@ccall` macro. In the future, Clang.jl should handle this, but for now, they're handled in `epilogue.jl`.

Gnimuc:
> We also pinned `CEnum` to an older version (for no good reason except that the we didn't want to change the generated code) and later _vendored_ the module.

As SCIP doesn't use duplicated enum members, so I just completely remove the use of `CEnum.jl`. Clang.jl's new generator now has an option for this.

rschwarz:
OK, so `LibSCIP.jl` is now the only file generated automatically by the script?
And the manual adjustments are all covered in `gen/prologue.jl`, `gen/epilogue.jl` or `src/wrapper.jl`?

That's a great improvement and will help a lot to keep up with API changes in SCIP in the future!

matbesancon:
this is fantastic yes

gen/generator.toml
```diff
@@ -0,0 +1,18 @@
...
+printer_blacklist = [
+    "printError",
+    "printErrorHeader"
+]
+
+[codegen]
+opaque_as_mutable_struct = false
+
+[codegen.macro]
+macro_mode = "basic"
```
matbesancon, line 18:
```suggestion
macro_mode = "basic"

```

gen/generator.toml
```diff
@@ -0,0 +1,18 @@
...
+library_name = "libscip"
+output_file_path = "../src/LibSCIP.jl"
+module_name = "LibSCIP"
+prologue_file_path = "./prologue.jl"
+epilogue_file_path = "./epilogue.jl"
+export_symbol_prefixes = ["SCIP_", "SCIP", "BMS_"]
+use_julia_native_enum_type = true
+printer_blacklist = [
+    "printError",
+    "printErrorHeader"
```
matbesancon, line 11:
```suggestion
    "printErrorHeader",
```

Gnimuc:
@matbesancon @rschwarz This is ready for another round of review. As for the test failures, I guess I didn't something wrong when solving the conflict. Feel free to edit this PR if needed. I'm going to merge the Clang.jl#278 this weekend and you could run the generator on your own machine.

Gnimuc:
I won't keep updating this PR but will keep monitoring the compatibility of the generated code [here](https://github.com/Gnimuc/GeneratorScripts/tree/master/S/SCIP).

matbesancon:
thanks for the work on this! So the current PR is based on what is currently on Clang.jl master right?

Gnimuc:
Yes. It's recommended to use Clang.jl#master from now on.

matbesancon:
there seems to be some conversion error: `unsafe_convert(::Type{Ptr{Ptr{Int8}}}, ::Base.RefValue{Cstring})`

Gnimuc:
This has been fixed in https://github.com/scipopt/SCIP.jl/pull/186/commits/eba7eff535a5dab3d36d2793f9820b77ba6b9816, but I did something wrong when solving the conflict, so it's not in the latest commit. I believe you could fix this by cherry-pick that commit or just simply fix it with a new one.

matbesancon:
@Gnimuc do you know when a release will be tagged on Clang.jl? Depending on master is a bit dangerous as next commits can break things

Gnimuc:
This package is being monitored in https://github.com/Gnimuc/GeneratorScripts, it's safe to just use the master branch.

Gnimuc:
Actually, this PR bundles the Modifest.toml, so it will always use the same version of Clang.jl unless someone does a `pkg> up`.

matbesancon:
@Gnimuc should we bump the gen/Project.toml to the latest 0.11 you released?

Gnimuc:
> @Gnimuc should we bump the gen/Project.toml to the latest 0.11 you released?

0.11? you mean 0.14? No, 0.14 has not been officially released yet. The latest master has some new features like variadic function support, docstrings dumping, etc. If you'd like to update to the latest master, run `pkg> up` in the project mode.

matbesancon:
ok thank you, we do not need this for now so merging

PR #189, opened 2021-03-26
SC changed to SCIP_CALL

etadewal:
Small change to syntax to be more in line with SCIP.


matbesancon (approved):

matbesancon:
Looks good. As it was an internal change I wouldn't consider this as breaking

PR #190, opened 2021-03-30
Moving ManagingSCIP to SCIPData

etadewal:


src/MOI_wrapper.jl
```diff
@@ -36,8 +38,18 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
...
+        @assert scip[] != C_NULL
+        @SCIP_CALL SCIPincludeDefaultPlugins(scip[])
+        @SCIP_CALL SCIP.SCIPcreateProbBasic(scip[], "")
+
+        scip_data = SCIPData(scip, Dict(), Dict(), 0, 0, Dict(), Dict(), Dict())
+
+        o = new(scip_data, PtrMap(), ConsTypeMap(), Dict(), Dict(), Dict(),
+        Nothing)   # Is scip_data the right argument here? in the original, new called SCIPData(), which is now included in the Optimizer function
+        finalizer(free_scip, o)
+    end
```
matbesancon, line 52: There shouldn't be an end here

src/conshdlr.jl
```diff
@@ -47,7 +47,7 @@ callback functions.
 From SCIP's point-of-view, this objects corresponds to the SCIP_CONSHDLRDATA,
 but its memory is managed by Julia's GC.
 
-It's recommended to store a reference to your instance of `ManagedSCIP` or
+It's recommended to store a reference to your instance of `SCIPData` or
 `SCIP.Optimizer` here, so that you can use it within your callback methods.
```
matbesancon, line 51:
```suggestion
It's recommended to store a reference to your instance of `SCIP.Optimizer`
here so that you can use it within your callback methods.
```

src/MOI_wrapper.jl
matbesancon: regarding the comment, new creates a new Optimizer, not a new SCIPData

etadewal:
Remaining TODO:
adjust finalizer for optimizer
rename variables (managed_scip, mscip and others that may be confusing now)

etadewal:
finalizer seems to work now
mscip has been changed to scipd (SCIPData) or o (Optimizer) everywhere, all calls to it have been revised as well and seem to work properly
67 tests currently succeed, however 25 errors encountered for "print statistics"

src/MOI_wrapper.jl
```diff
@@ -26,18 +26,29 @@ const VEC_TERM = MOI.VectorAffineTerm{Float64}
...
 mutable struct Optimizer <: MOI.AbstractOptimizer
-    mscip::ManagedSCIP
+    inner::SCIPData
     reference::PtrMap
     constypes::ConsTypeMap
     binbounds::Dict{VI,BOUNDS} # only for binary variables
     params::Dict{String,Any}
     start::Dict{VI,Float64} # can be partial
-    moi_separator::Any #Union{CutCbSeparator, Nothing}
+    moi_separator # ::Union{CutCbSeparator, Nothing}
```
matbesancon, line 38: the type annotation should be readded

src/MOI_wrapper.jl
```diff
@@ -26,18 +26,29 @@ const VEC_TERM = MOI.VectorAffineTerm{Float64}
...
+        scip = Ref{Ptr{SCIP_}}(C_NULL)
+        @SCIP_CALL SCIPcreate(scip)
+        @assert scip[] != C_NULL
+        @SCIP_CALL SCIPincludeDefaultPlugins(scip[])
+        @SCIP_CALL SCIP.SCIPcreateProbBasic(scip[], "")
+
+        scip_data = SCIPData(scip, Dict(), Dict(), 0, 0, Dict(), Dict(), Dict())
+
+        o = new(scip_data, PtrMap(), ConsTypeMap(), Dict(), Dict(), Dict(),
+        Nothing)   # Is scip_data the right argument here? in the original, new called SCIPData(), which is now included in the Optimizer function
```
matbesancon, line 50:
```suggestion
        o = new(scip_data, PtrMap(), ConsTypeMap(), Dict(), Dict(), Dict(), nothing)
```

src/MOI_wrapper.jl
```diff
@@ -152,7 +181,7 @@ function MOI.get(o::Optimizer, ::MOI.TimeLimitSec)
 end
 
 function MOI.set(o::Optimizer, ::MOI.TimeLimitSec, value)
-    if value === nothing
+    if value == nothing
```
matbesancon, line 184: why did you change this? `===` should be preferred when comparing with nothing

matbesancon, line 184:
```suggestion
    if value === nothing
```

src/MOI_wrapper.jl
etadewal: yes I just reinserted it, don't know why I removed it in the first place

src/MOI_wrapper.jl
etadewal: I thought I made a typo here, have restored the original === now

matbesancon:
@rschwarz a recap on what this PR is about:
removing more or less the layer of logic that was on ManagedSCIP.
To avoid adding too many fields, the pointers are still bundled in SCIPData, but the memory management happens at the Optimizer struct level, similarly to what @odow did in HiGHS.jl

rschwarz:
That approach makes sense to me and will be a net simplification of the code :+1: 

If the memory management is moved out of `ManagedSCIP` (now `SCIPData`), this means that this type can no longer be effectively used on its own, only through the `Optimizer` wrapper, right? Maybe this should be emphasized with a comment. The README also has a paragraph on this that might benefit from an update.

It looks to me that this set of changes eliminates the additional layer mostly on the data type side, but not for the functions. Is it also planned to remove the intermediate functions, such as [add_linear_constraint](https://github.com/scipopt/SCIP.jl/blob/0d0321d31ae177a71228c8055274ebbd047b40ca/src/scip_data.jl#L191). I think these functions would no longer be used, because the user will just use the `MOI.add_constraint`. Similarly for much code in `MOI_wrapper/` that calls such intermediate functions, with some logic for array conversion etc.

With this `ManagedSCIP` layer gone, one could also remove the non-MOI tests, potentially migrating some of them to MOI-based tests.

These suggestions could also be done in a later PR (or ignored), of course.

rschwarz:
Another simplification that would be possible (by elimination of the `ManagedSCIP` layer) is the elimination of the [VarRef](https://github.com/scipopt/SCIP.jl/blob/0d0321d31ae177a71228c8055274ebbd047b40ca/src/scip_data.jl#L2) and `ConsRef` types. Here, we should simply use `MOI.VariableIndex` etc.

These types were only created to avoid dependency on `MOI` at this layer. See [add_variable](https://github.com/scipopt/SCIP.jl/blob/0d0321d31ae177a71228c8055274ebbd047b40ca/src/MOI_wrapper/variable.jl#L3) for how the MOI wrapper converts between the SCIP.jl and MOI types (using the same inner integer value).

Further, the [var](https://github.com/scipopt/SCIP.jl/blob/0d0321d31ae177a71228c8055274ebbd047b40ca/src/MOI_wrapper.jl#L86) and `cons` utilities would then be simplified.

README.md
```diff
@@ -75,14 +75,14 @@ Julia). Convenience wrapper functions based on Julia types are added as needed.
 **Memory Management**: Programming with SCIP requires dealing with variable and
 constraints objects that use [reference
 counting](https://scip.zib.de/doc-6.0.0/html/OBJ.php) for memory management.
-SCIP.jl provides a wrapper type `ManagedSCIP` that collects lists of `SCIP_VAR*`
+SCIP.jl provides a wrapper type `SCIPData` that collects lists of `SCIP_VAR*`
```
rschwarz, line 78: Maybe it's better to explain that this type is an implementation detail and that it should not be used by end users directly. Similar to the functions mentioned below (`add_variable` etc).

src/MOI_wrapper.jl
```diff
@@ -50,22 +59,40 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
...
+    # Avoid double-free (SCIP will set the pointers to NULL).
+    if o.inner.scip[] != C_NULL
+        for c in values(o.inner.conss)
+            @SCIP_CALL SCIPreleaseCons(o.inner, c)
+        end
+        for v in values(o.inner.vars)
+            @SCIP_CALL SCIPreleaseVar(o.inner, v)
+        end
+        # only o.inner.scip is GC-protected during ccall!
+        GC.@preserve o begin
```
rschwarz, line 73: I'm not sure whether this is correct. Maybe the `GC.@preserve` is not needed at all, because this function already has `o` as an argument. It could also be that `o.inner` should be passed to the `GC.@preserve`. I forgot the specific need for this macro call.

src/scip_data.jl
```diff
@@ -8,8 +8,9 @@ struct ConsRef
     val::Int64
 end
 
-"ManagedSCIP holds pointers to SCIP data and takes care of memory management."
-mutable struct ManagedSCIP
+#to be moved to MOI_wrapper
```
rschwarz, line 11: Should this still be done, or this comment removed?

More consistent with the idea of this PR would be to move the content of `MOI_wrapper/` into the parent directory and eliminate many of the existing source files there, right?

src/scip_data.jl
```diff
@@ -26,50 +27,41 @@ mutable struct ManagedSCIP
...
-        @SCIP_CALL SCIP.SCIPcreateProbBasic(scip[], "")
-
-        mscip = new(scip, Dict(), Dict(), 0, 0, Dict(), Dict(), Dict())
-        finalizer(free_scip, mscip)
-    end
 end
 
-# Protect ManagedSCIP from GC for ccall with Ptr{SCIP_} argument.
-Base.unsafe_convert(::Type{Ptr{SCIP_}}, mscip::ManagedSCIP) = mscip.scip[]
+# TODO: all references to scipd / SCIPData need to be changed to (reference to optimmizer struct?) / SCIPData
```
rschwarz, line 32: This is done, so the TODO comment may be dropped, right?

rschwarz (commented):
Not sure whether my proposal is within scope of this PR, but if the goal is to eliminate the intermediate layer (for the purpose of simplification), and keep `SCIPData` only to group the pointer data, then we should also try to eliminate all functions that take `scipd::SCIPData` as (first) argument, rather than `o::Optimizer`.

src/MOI_wrapper.jl
```diff
@@ -50,22 +59,40 @@ mutable struct Optimizer <: MOI.AbstractOptimizer
...
+            @SCIP_CALL SCIPfree(o.inner.scip)
+        end
+    end
+    @assert o.inner.scip[] == C_NULL
+end
+
 # Protect Optimizer from GC for ccall with Ptr{SCIP_} argument.
-Base.unsafe_convert(::Type{Ptr{SCIP_}}, o::Optimizer) = o.mscip.scip[]
+Base.unsafe_convert(::Type{Ptr{SCIP_}}, o::Optimizer) = o.inner.scip[]
```
odow, line 81: Please read https://jump.dev/MathOptInterface.jl/stable/manual/implementing/#Optimizer-objects-for-C-solvers

src/MOI_wrapper.jl
odow: You should implement `cconvert` for the type instead of using `@preserve` https://jump.dev/MathOptInterface.jl/stable/manual/implementing/#Optimizer-objects-for-C-solvers

odow (commented):
The approach I have taken that has worked very well in other wrappers is to have absolutely no intermediate layers. The `Optimizer` stores the raw pointers. No functions like `add_linear_constraint` are exported. There is only the C API and the MOI API.

src/MOI_wrapper.jl
rschwarz: Thanks for the clarification. There's already an implementation of [unsafe_convert](https://github.com/scipopt/SCIP.jl/blob/0d0321d31ae177a71228c8055274ebbd047b40ca/src/scip_data.jl#L35) for `SCIPData`.

src/MOI_wrapper.jl
rschwarz: Note that the `[]` operator is used here, because the [SCIPfree](https://www.scipopt.org/doc-7.0.2/html/group__GeneralSCIPMethods.php#gad8ae80418f738b5b10e3b0fd08782bed) function expects and argument `SCIP ** scip` (which will be set to `NULL`). That is, we don't store `SCIP*` but actually `SCIP**`.

etadewal:
> Not sure whether my proposal is within scope of this PR, but if the goal is to eliminate the intermediate layer (for the purpose of simplification), and keep `SCIPData` only to group the pointer data, then we should also try to eliminate all functions that take `scipd::SCIPData` as (first) argument, rather than `o::Optimizer`.

I've tried this and got 
`ERROR: LoadError: LoadError: UndefVarError: Optimizer not defined
Stacktrace:
 [1] top-level scope at /home/erik/scipjl/SCIP.jl/src/conshdlr.jl:299
 [2] include(::Function, ::Module, ::String) at ./Base.jl:380
 [3] include at ./Base.jl:368 [inlined]
 [4] include(::String) at /home/erik/scipjl/SCIP.jl/src/SCIP.jl:1
 [5] top-level scope at /home/erik/scipjl/SCIP.jl/src/SCIP.jl:19
 [6] include(::Function, ::Module, ::String) at ./Base.jl:380
 [7] include(::Module, ::String) at ./Base.jl:368
 [8] top-level scope at none:2
 [9] eval at ./boot.jl:331 [inlined]
 [10] eval(::Expr) at ./client.jl:467
 [11] top-level scope at ./none:3
in expression starting at /home/erik/scipjl/SCIP.jl/src/conshdlr.jl:299
in expression starting at /home/erik/scipjl/SCIP.jl/src/SCIP.jl:19
ERROR: LoadError: Failed to precompile SCIP [82193955-e24f-5292-bf16-6f2c5261a85f] to /home/erik/.julia/compiled/v1.5/SCIP/a0EUw_cevbp.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1305
 [3] _require(::Base.PkgId) at ./loading.jl:1030
 [4] require(::Base.PkgId) at ./loading.jl:928
 [5] require(::Module, ::Symbol) at ./loading.jl:923
 [6] include(::String) at ./client.jl:457
 [7] top-level scope at none:6
in expression starting at /home/erik/scipjl/SCIP.jl/test/runtests.jl:2
ERROR: Package SCIP errored during testing
`

src/MOI_wrapper.jl
etadewal: I've removed the `@preserve` in line 73 and added `Base.cconvert(::Type{Ptr{SCIP_}}, o::Optimizer) = o` right before the `unsafe_convert` in line 81 but I don't think that's quite correct. Where have I gone wrong?

src/scip_data.jl
etadewal: Yes, the line has been deleted. Thanks for pointing it out.

src/MOI_wrapper.jl
matbesancon: > but I don't think that's quite correct

what do you mean specifically?

src/scip_data.jl
etadewal: I think this was a comment from when I started to map out what should be moved where, since I thought it would be necessary. Since the current build works, I wouldn't move this to MOI_wrapper unless anybody with more insight than me wants me to (or to alternatively move the contents of MOI_wrspper/ into the parent dir).

src/MOI_wrapper.jl
etadewal: I'm getting 

> error in running finalizer: AssertionError(msg="o.inner.scip[] == C_NULL")
create and semi-manual free: Test Failed at /home/erik/scipjl/SCIP.jl/test/scip_data.jl:26
  Expression: o.inner.scip[] == C_NULL
   Evaluated: Ptr{Nothing} @0x00000000024680c0 == Ptr{Nothing} @0x0000000000000000
Stacktrace:
 [1] top-level scope at /home/erik/scipjl/SCIP.jl/test/scip_data.jl:26
 [2] top-level scope at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.5/Test/src/Test.jl:1115
 [3] top-level scope at /home/erik/scipjl/SCIP.jl/test/scip_data.jl:23

etadewal: Found it, I misplaced a line.

matbesancon:
@etadewal you need to include MOI_wrapper.jl before conshdlr.jl in the SCIP.jl file

README.md
```diff
@@ -75,14 +75,15 @@ Julia). Convenience wrapper functions based on Julia types are added as needed.
...
-can be used to retrieve the `SCIP_VAR*` or `SCIP_CONS*` pointer via `get_var`
-and `get_cons` respectively.
+collected itself (via `finalize`). `SCIPData` is an internal structure used by
+`Optimizer`, which is what should be used preferably. When adding a variable 
+(`add_variable`) or a constraint (`add_linear_constraint`), an integer index
+is returned. This index can be used to retrieve the `SCIP_VAR*` or `SCIP_CONS*`
+pointer via `get_var` and `get_cons` respectively.
 
-`ManagedSCIP` does not currently support deletion of variables or constraints.
+`SCIPData` does not currently support deletion of variables or constraints.
```
matbesancon, line 86: This should be removed, we'll now consider SCIPData an internal thing that doesn't need to be in the user documentation

etadewal: Line has been removed in the last commit

PR #203, opened 2021-09-08
upgrade to MOI 10

matbesancon:
Still some tests to fix, some probably been move around.


matbesancon:
Some tests are a bit weird to fix (what is the solver supposed to do when a BadModel cannot be copied? there is no docstring)
`duplicate_VariableName` feels like it should not be run when VariableName was excluded.

matbesancon:
Still some strange things with indicator constraints:
```
#     test_constraint_Indicator_ACTIVATE_ON_ZERO: Error During Test at /home/mbesancon/.julia/packages/MathOptInterface/Evu1m/src/Test/test_constraint.jl:868
#   Test threw exception
#   Expression: MOI.get(model, MOI.ConstraintSet(), c) == s
#   ArgumentError: Bridge of type `MathOptInterface.Bridges.Constraint.IndicatorActiveOnFalseBridge{Float64, MathOptInterface.VectorAffineFunction{Float64}, MathOptInterface.GreaterThan{Float64}}` does not support accessing the attribute `MathOptInterface.ConstraintSet()`. If you encountered this error unexpectedly, it probably means your model has been reformulated using the bridge, and you are attempting to query an attribute that we haven't implemented yet for this bridge. Please open an issue at https://github.com/jump-dev/MathOptInterface.jl/issues/new and provide a reproducible example explaining what you were trying to do.
```
to check later

matbesancon:
OK we are getting close to a mergeable state, I removed the bound tests for now

matbesancon:
@odow in the two commits starting with "remove bound" above, I am not sure if this was the intended behavior or not.

matbesancon:
I'd like to see if there are some containers we can simplify, if feels like we have a ton of dictionaries storing various constraint and variable indices, never sure which one should be checked

src/MOI_wrapper.jl
```diff
@@ -135,16 +133,23 @@ end
 
 MOI.get(::Optimizer, ::MOI.SolverName) = "SCIP"
 
-MOIU.supports_default_copy_to(model::Optimizer, copy_names::Bool) = !copy_names
+MOIU.supports_default_copy_to(::Optimizer, copy_names::Bool) = !copy_names
```
odow, line 136:
```suggestion
MOI.supports_incremental_interface(::Optimizer) = true
```

src/MOI_wrapper.jl
```diff
@@ -214,7 +219,7 @@ function MOI.empty!(o::Optimizer)
 end
 
 function MOI.copy_to(dest::Optimizer, src::MOI.ModelLike; kws...)
```
odow, line 221:
```suggestion
function MOI.copy_to(dest::Optimizer, src::MOI.ModelLike)
```

src/MOI_wrapper.jl
```diff
@@ -214,7 +219,7 @@ function MOI.empty!(o::Optimizer)
...
-    return MOIU.automatic_copy_to(dest, src; kws...)
+    return MOIU.default_copy_to(dest, src; kws...)
```
odow, line 222:
```suggestion
    return MOIU.default_copy_to(dest, src)
```

test/MOI_wrapper_bridged.jl
```diff
@@ -4,60 +4,23 @@ const MOIB = MOI.Bridges
...
+const CONFIG_BRIDGED = MOIT.Config(atol=1e-5, rtol=1e-5, exclude=Any[
+    MOI.ConstraintDual, MOI.ConstraintName, MOI.VariableName, MOI.DualObjectiveValue, MOI.VariableBasisStatus, MOI.ConstraintBasisStatus,
+])
+
+@testset "MOI unit tests" begin
+    excluded = copy(MOI_BASE_EXCLUDED)
+    append!(excluded, [
+        "test_linear_Interval_inactive",
+        "test_linear_integration",
+        "test_model_ordered_indices", # TODO should fix? ListOf in order of creation
```
odow, line 16: Yes, this needs to be fixed.

I usually add a comment why we're excluding tests as well.  `test_linear_integration` is a fairly important test, so it'd be nice to know why it's skipped.

matbesancon: fixed here. I added an explanation for the integration test, deleting a variable in SCIP is only allowed in certain stages so we are fairly conservative in the wrapper

rschwarz:
> I'd like to see if there are some containers we can simplify, if feels like we have a ton of dictionaries storing various constraint and variable indices, never sure which one should be checked

Not sure what you mean. In SCIPData, there's only one `Dict` each for variables and constraints.

In the `Optimizer` struct, there are more dictionaries, but these store different values (e.g. bounds, starting point).

matbesancon:
yes the confusion was more in some parts on what dict is supposed to be used for checking that something exists

PR #213, opened 2021-11-16
MINLP test re-integration

matbesancon:
I expect some things to break as we do.

The goals are:
- to fix the test call structure to make it compatible with JuMP 0.22
- to re-add a github action testing MINLP


matbesancon:
tackles https://github.com/scipopt/SCIP.jl/issues/206

PR #218, opened 2021-12-09
Allow stage PRESOLVING in assert_solved

rschwarz:
This might happen if a user limit (ie time) stops the solver during presolving.

Fixes a bug [encountered](https://github.com/NREL-SIIP/PowerSimulations.jl/runs/4467002116?check_suite_focus=true#step:5:316) by @jd-lara.


matbesancon:
looks good! I don't think there is a good way to test this anyways, except having a long-to-presolve instance which we don't want on CI

jd-lara:
FWIW we trigger this because in our CI we have a solve/time limit of a relatively large model.

PR #231, opened 2022-04-29
prep scip 8

matbesancon:
Preparation for SCIP 8 release of the wrapper


fschloesser:
Looks okay to me, although i don't know much about julia ;)

PR #239, opened 2022-08-15
Throw DeleteNotAllowed error when deleting variables

odow:
In addition, greatly simplify the MOI_wrapper_cached.jl file.

Thowing a `DeleteNotAllowed` will trigger the cache to rebuild the model. This means we can now run the full gamut of tests when using the cache. 

I don't know if there is value in running `_cached`, `_bridged`, and `_direct`. Perhaps we could delete the others? We really only care that the cache is working, and the number of excludes suggests that the rest of the API isn't fully implemented in direct mode.

### DIFF
--- a/src/MOI_wrapper/variable.jl
+++ b/src/MOI_wrapper/variable.jl
@@ -28,7 +28,12 @@ function MOI.delete(o::Optimizer, vi::VI)
     # the variable-in-constraint relation, so, to be conservative, we only allow
     # to delete a variable when there are no constraints in the model.
     if length(o.inner.conss) > 0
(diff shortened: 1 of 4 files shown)


matbesancon:
thanks a lot that's great!

> We really only care that the cache is working, and the number of excludes suggests that the rest of the API isn't fully implemented in direct mode

there is value in having a direct mode since the solver can, for rapid model building, modification and solve loops

odow:
I didn't actually run all the tests locally, so I have a couple of fixes about to be incoming.

odow:
Okay. Passing now. The cpsat test will be fixed in MOI v1.7

PR #241, opened 2022-08-26
Cache result of SCIPgetSols

matbesancon:
instead of always refetching all solutions, creates and stores the vector once after solve


odow:
@matbesancon still want this?

odow:
I think I'm going to close this as won't fix. #342 has most of the improvement. There's probably plenty of room in the wider SCIP.jl for a benchmark and performance improvements, but we can leave that until someone complains.

PR #251, opened 2022-11-02
relaxation handler

matbesancon:
expose a relaxation handler in the wrapper with an idiomatic interface

### DIFF
--- a/src/MOI_wrapper/conshdlr.jl
+++ b/src/MOI_wrapper/conshdlr.jl
@@ -71,3 +71,7 @@ end
 function include_cutsel(o::Optimizer, cutsel::CS; name = "", description = "", priority=10000) where {CS <: AbstractCutSelector}
     return include_cutsel(o.inner.scip[], cutsel, o.inner.cutsel_storage; name=name, description=description, priority=priority)
 end
+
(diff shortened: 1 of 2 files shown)


rschwarz:
How do you imagine the relaxation handler would look at the problem? Through the MOI representation, available via `Optimizer`, or on a lower level, with `SCIPData`?

I'm amazed how little code is present in this PR. Do you imagine that constraint handlers could've also be implemented in such a simple manner?

matbesancon:
> How do you imagine the relaxation handler would look at the problem? Through the MOI representation, available via Optimizer, or on a lower level, with SCIPData?

That's a good question, I don't think it's super nice to force people to handle low-level stuff all the time but we can't cover all the possible things they could try from there. So the first step, just like cut selection, will be to have a simple-enough interface function
(compute and return a lower bound).

One main problem is having to work on the transformed problem itself, so not variables that correspond to the MOI-indexed ones

rschwarz:
I've never implemented a relaxation handler, but one would also need some information about the tree node. I guess that this is given implicitly via variable bounds in the (local) transformed problem, so that there really is no alternative to looking at this.

matbesancon:
yes that's my fear. So people would have to look at the (very sparse) SCIP documentation on relaxation handlers. The good thing is that it is at least accessible (without having to write a C-function)

PR #255, opened 2023-01-24
Fix return value of ListOfModelAttributesSet

odow:
x-ref: https://github.com/jump-dev/MathOptInterface.jl/pull/2085

There are actually a bunch of excluded tests in SCIP. That's usually suggestive of bugs in the implementation.


matbesancon:
thanks a lot, yes indeed I did not implement the behavior according to the MOI standard, not a bug per se but this is a very peculiar expectation and I was avoiding to move to far away from the solver behaviour

matbesancon:
but more coverage and consistency is a positive thing

PR #267, opened 2023-05-04
use original variables in get()

hannahtro:
- solves https://github.com/scipopt/SCIP.jl/issues/257#
- use `SCIPvarGetOrigvarSum` to use original variables instead of transformed ones in `MOI.get()`


src/MOI_wrapper/indicator_constraints.jl
```diff
@@ -53,10 +53,22 @@ function MOI.get(
...
+    scalar = Ref(1.0)
+    constant = Ref(0.0)
+    orig_vars = []
+    for i in 1:nvars
+        @assert scalar.x == 1.0 
+        @assert constant.x == 0.0
+        var = Ref(vars[i])
+        @SCIP_CALL SCIPvarGetOrigvarSum(var, scalar, constant)
+        push!(orig_vars, var[])
+    end
```
matbesancon, line 66:
```suggestion
    orig_vars = map(1:nvars) do i
        var = Ref(vars[i])
        @SCIP_CALL SCIPvarGetOrigvarSum(var, scalar, constant)
        @assert scalar[] == 1.0 
        @assert constant[] == 0.0
        var[]
    end
```

Given that the code is repeated, might be worth putting in a separate function and call it at the two places

src/MOI_wrapper/variable.jl
```diff
@@ -389,3 +389,16 @@ end
...
+    constant = Ref(0.0)
+    orig_vars = map(1:nvars) do i
+        var = Ref(vars[i])
+        @SCIP_CALL SCIPvarGetOrigvarSum(var, scalar, constant)
+        @assert scalar[] == 1.0 
+        @assert constant[] == 0.0
+        var[]
+    end
+    return orig_vars
+end
```
matbesancon, line 404:
```suggestion
end

```

PR #273, opened 2023-08-18
Update SCIP.jl to support Ipopt precompiled with LBT

amontoison:
`SCIP_PaPILO_jll` is precompiled with an upper bound on `Ipopt_jll`. We can't download the last artifact of `Ipopt_jll` when we use Ipopt.jl if SCIP.jl is installed.
I updated SCIP.jl such that a BLAS / LAPACK backend is automatically loaded. You should be able to drop this [constraint](https://github.com/JuliaPackaging/Yggdrasil/blob/master/S/SCIP_PaPILO/build_tarballs.jl#L71) in the future.


matbesancon:
Perfect thanks! We'll be able to drop the constraint with the new bugfix release in the coming week

PR #278, opened 2023-10-26
Add license header to .jl files

odow:


src/LibSCIP.jl
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2018 Felipe Serrano, Miles Lubin, Robert Schwarz, and contributors
```
matbesancon, line 1: we shouldn't add a header here since this is a generated file

odow: Removed (also from `gen/prologue.jl`)

PR #279, opened 2023-10-26
WIP: Add JuliaFormatter and remove acronyms

odow:
Should be rebased after #278

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -3,32 +3,14 @@
 # Use of this source code is governed by an MIT-style license that can be found
 # in the LICENSE.md file or at https://opensource.org/licenses/MIT.
 
-using MathOptInterface
-const MOI = MathOptInterface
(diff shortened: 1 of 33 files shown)


src/MOI_wrapper.jl
```diff
@@ -1,29 +1,16 @@
-using MathOptInterface
-const MOI = MathOptInterface
-const MOIU = MOI.Utilities
```
matbesancon, line 3: MOIU is standard no?

odow: I really don't like it. `MOI.Utilities` is much clearer, and you very rarely need to use `MOIU` anyway.

odow:
One downside to this PR is that it will screw up all the open PRs. So I'll rebase, but don't feel pressured to merge

matbesancon:
that's fine, all of them are inactive or just trials that will not be merged for now

Issue #286, opened 2024-04-04
Update to allow SCIP version 9?


GregPlowman:
Can SCIP.jl be updated to allow SCIP version 9.

It seems there are version 9 `jll`s for SCIP and SCIP_PaPILO:
```
SCIP_jll 900.0.0+1
SCIP_PaPILO_jll 900.0.0+0
```

Currently, SCIP version is limited in `init.jl`:
```
current = VersionNumber("$major.$minor.$patch")
required = VersionNumber("8")
upperbound = VersionNumber("9")
if current < required || current >= upperbound
    @error(
        "SCIP is installed at version $current, " *
        "supported are $required up to (excluding) $upperbound."
    )
end
```
Note that `current >= upperbound` results in the error.
Manually editing to `current > upperbound` seems to work with a manual installation of SCIP version 9, which is required for Windows.

matbesancon:
fixed in recent PR

PR #289, opened 2024-05-07
Minimal changes to support SCIP 9

hellemo:
Minimal necessary changes to test with SCIP 9. All tests pass locally, but there seems to be an issue in CI with julia 1.6.

#286

### DIFF
--- a/.github/workflows/MINLP.yml
+++ b/.github/workflows/MINLP.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/MINLP.yml
```diff
@@ -11,12 +11,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        version: ['1.6', '1']  # Test against current minor release (at least 1.3)
+        version: ['lts', '1']  # Test against current minor release (at least 1.3)
```
matbesancon, line 14:
```suggestion
        version: ['lts', '1']
```

PR #306, opened 2025-05-02
Fix modification when stage is left in SCIP_STAGE_SOLVING

odow:
Closes https://github.com/scipopt/SCIP.jl/issues/304

### DIFF
--- a/src/MOI_wrapper.jl
+++ b/src/MOI_wrapper.jl
@@ -146,7 +146,7 @@ end
 
 "Go back from solved stage to problem modification stage, invalidating results."
 function allow_modification(o::Optimizer)
-    if !(SCIPgetStage(o) in (SCIP_STAGE_PROBLEM, SCIP_STAGE_SOLVING))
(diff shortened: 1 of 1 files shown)


odow:
The segfault is because constraint handlers can call MOI code **during the solve**:
https://github.com/scipopt/SCIP.jl/blob/23d65818215543d6a42ed09f1adb01425ba3bfc8/test/conshdlr_support.jl#L260-L267
that seems pretty weird to me

odow:
As a related side benefit: this fix also allows us to fix https://github.com/scipopt/SCIP.jl/issues/285, because now we can tell when `SCIP_STATUS_UNKNOWN` corresponds to `MOI.OPTIMIZE_NOT_CALLED` and when it corresponds to `MOI.OTHER_ERROR`.

matbesancon:
Yes there is no distinction in the SCIP dynamic mechanisms between a constraint added before the solve started and one during the solve normally, it's all "add constraints" in any case

PR #333, opened 2025-05-14
Add CODECOV_TOKEN

odow:


odow:
@matbesancon I don't have permission to add secrets. Can you help?

Here are the docs: https://docs.codecov.com/docs/adding-the-codecov-token

matbesancon:
Done! Should work now

PR #340, opened 2025-05-16
Update src/LibSCIP using the latest SCIP_jll

odow:
So one thing I realized but didn't realize: woah, that's a pretty broad API.

Not sure if it's worth merging this. Or if we can slim down some of the API that we're covering?


odow:
I'm going to leave this as-is for now.

PR #353, opened 2026-02-20
SCIP IIS with MOI Conflict interface

matbesancon:
Adding the new IIS feature to compute conflicts.
For now this is quite a hacky path, relying on variable and constraint names to verify whether a constraint or variable is present in the IIS or not
cc @Opt-Mucca 

I haven't found any better way to match the IIS sub-SCIP information to the original model so far, I don't think the IIS produces a correspondence map between original and IIS variables / constraints


matbesancon:
without surprise, CI cannot pass at the moment because of the finding-by-name issue. Depending on how easy it can be worked out, we'll either copy / adapt the test setup by adding names

Opt-Mucca:
@matbesancon There is a map from the original problem to the IIS subscip. See the struct `SCIP_IIS` with `varsmap` and `conssmap` in `struct_iisfinder`. That can be used to determine which constraints are in the IIS (I don't think I programmed something in the reverse direction though, i.e., from subscip to original SCIP).

matbesancon:
unless I missed it, I don't see any methods to access these two hashmaps

Opt-Mucca:
That was probably laziness on my part......

matbesancon:
tss tsss
Then we should merge one of the two MRs I opened on SCIP to expose the necessary information. Feel free to chime in on your favorite version

